PR #2, opened 2018-06-10
fix: corrections on readme

diegohaz:
Just fixed some things as I read.


pedronauck:
Thanks @diegohaz, to be honest I didn't release the project yet but I appreciate a lot your early review... 😂

PR #7, opened 2018-06-12
Correct Wepack to Webpack

aaronang:


diegohaz:
Also, according to [this](https://github.com/webpack/media#name),
> webpack should always be written in lower-case letters. Even at the beginning of a sentence.

thompsonemerson (approved):

pedronauck:
I agreed @diegohaz ✌️ Can you make this little change @aaronang ?

aaronang:
Sorry for my late reply. Will do 👍

pedronauck (approved):

PR #60, opened 2018-06-19
Fix documentation for babel plugin

Swapnull:
### Description

Add here a description about your Pull Request

The README for docz-plugin-babel6 currently wont work. 

It currently says
```javascript
import { babel } from 'docz-plugin-babel6'
export default { plugins[babel6()] };
```
But this will result in 'babel6 is not defined' 

In this PR I have changed the export name from the plugin to be babel6 as this makes most sense to me. 
It would possibly be safer, but may make less sense, to just update the README to either

- change the import to `import { babel as babel6 }`
- keep the current import and do `plugins[babel()]`, which is less descriptive. 

### Review

- [ ] Test out the new plugin

### DIFF
--- a/packages/docz-plugin-babel6/README.md
+++ b/packages/docz-plugin-babel6/README.md
@@ -3,7 +3,7 @@
 By default and by some performance issues, Docz use Babel@7. If you want to use older versions of babel import this plugin and use on your `doczrc.js`:
 
 ```js
-import { babel } from 'docz-plugin-babel6'
+import { babel6 } from 'docz-plugin-babel6'
 
 export default {
   plugins: [babel6()]
--- a/packages/docz-plugin-babel6/README.md
+++ b/packages/docz-plugin-babel6/README.md
@@ -3,9 +3,9 @@
 By default and by some performance issues, Docz use Babel@7. If you want to use older versions of babel import this plugin and use on your `doczrc.js`:
 
 ```js
-import { babel6 } from 'docz-plugin-babel6'
+import { babel } from 'docz-plugin-babel6'
 
 export default {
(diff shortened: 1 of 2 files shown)


pedronauck:
the plugin export a variable called `babel` @Swapnull, just change it, please 🙏

pedronauck:
sorry @Swapnull, maybe I didn't be so clear... but I had asked you to change to the correct value exported. Change the exported value now I'll result in a break change version, so I think that is better keep the value exported as `babel`. So, just change on readme to be correct, please :pray:

packages/docz-plugin-babel6/src/index.ts
```diff
@@ -9,7 +9,7 @@ const modifyHappypackLoader = (plugin: any) => {
   plugin.config.loaders[0].loader = require.resolve('babel-loader')
 }
 
-export const babel = () =>
```
pedronauck: please keep as `babel` 🙏

packages/docz-plugin-babel6/README.md
```diff
@@ -3,7 +3,7 @@
 By default and by some performance issues, Docz use Babel@7. If you want to use older versions of babel import this plugin and use on your `doczrc.js`:
 
 ```js
-import { babel } from 'docz-plugin-babel6'
```
pedronauck: keep this as `babel` too ✌️

pedronauck (changes requested):

Swapnull:
@pedronauck I have changed it back to exporting `babel` and updated the readme 😄

pedronauck (approved):

Issue #69, opened 2018-06-20
Config loader quietly falls back to defaults when it encounters an error


good-idea:
## Bug Report

**Describe the bug**

When `doczrc.js` or anything imported within it contains an error, it stops loading custom config and falls back to the defaults. 

*A clear and concise description of what the bug is.*

**To Reproduce**
in your `doczrc.js`:

```
console.log('Works')
const a = 1
a = 2

console.log('this will not print')

module.exports = {
        source: './src',
        debug: true,
}
```

**Expected behavior**

The command should throw the error and exit. Instead, docz continues to run, (presumably) using the default config.

good-idea:
Made a PR!

pedronauck:
## 🎉&nbsp; Fixed and released in the new [v0.4.0](https://github.com/pedronauck/docz/releases/tag/v0.4.0)

PR #70, opened 2018-06-20
Throws errors and exist when config encounters an error

good-idea:
Addresses #69 

### Description

Previously, the config loader would fall back to the default configuration when it encountered an error loading `doczrc.js`. This PR changes the behavior to report the error and fail instead.

### DIFF
--- a/packages/load-cfg/package.json
+++ b/packages/load-cfg/package.json
@@ -27,6 +27,7 @@
   },
   "devDependencies": {
     "@types/find-up": "^2.1.1",
-    "@types/node": "^10.3.2"
+    "@types/node": "^10.3.2",
+    "libundler": "^1.7.1"
   }
 }
--- a/packages/load-cfg/package.json
+++ b/packages/load-cfg/package.json
@@ -27,7 +27,6 @@
   },
   "devDependencies": {
     "@types/find-up": "^2.1.1",
-    "@types/node": "^10.3.2",
-    "libundler": "^1.7.1"
+    "@types/node": "^10.3.2"
   }
 }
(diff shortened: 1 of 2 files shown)


good-idea:
This also installs `libundler` as a devdependency - I don't like making global installs. I didn't add the `package-lock.json` that was generated - I wasn't sure if you had ignored this for a reason or not.

packages/load-cfg/src/index.ts
```diff
@@ -43,7 +43,8 @@ export const loadFile = (
       file = JSON.parse(fs.readFileSync(filepath, 'utf-8'))
     }
   } catch (err) {
-    file = defaultFile
+    console.warn('There was an error loading your doczrc.js config:')
```
pedronauck: please, remove `doczrc.js` because `load-cfg` is a generic package not specific!

pedronauck (changes requested):

packages/load-cfg/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "devDependencies": {
     "@types/find-up": "^2.1.1",
-    "@types/node": "^10.3.2"
+    "@types/node": "^10.3.2",
+    "libundler": "^1.7.1"
```
pedronauck: I don't need to add `libundler` as dependency, we're using lerna. With lerna you can have `devDependencies` just in the root package! But you need to run `lerna bootstrap`

good-idea: got it, thanks!

pedronauck (approved):

PR #90, opened 2018-06-26
Add overflow: auto to Sidebar

leesiongchan:
### Description

I can't see my components when it reached the maximum screen size height


marceloavf:
Thanks for the contribution @leesiongchan!

I think it's better use `overflow-y: auto;` and also add `margin-top: 64px;` to the sidebar `Footer` to keep the distance from top to bottom equal.

pedronauck:
Closed in favor of #103

PR #93, opened 2018-06-26
[docz-theme-default] Remove color transparent on body

bpetetot:
### Description

Fix applied to docz-theme-default package

My components are displayed with a transparent text (especially all components using portals).
It's due to the style `color: transparent` set on `body` element.

I don't know why it was set like that, but I propose to remove it to have by default a visible text color for components using portals.

I hope it helps. Docz is a great tool ;)


pedronauck:
Thanks @bpetetot 🙏

Issue #40, opened 2018-06-14
CLI it’s running on port 3000, but that port is already taken. (Actually running on 3001?)


filipdanic:
## Bug Report

The CLI is saying:
```
DONE  Compiled successfully in 166ms                                                                                                                                                                                                                              
You application is running at http://localhost:3000
```
But I already have an app running there. When I go to the port, my app is running there. Not `docz`. Meanwhile, there’s no indication of an error in the console.

But I tried port `3001` and it seems `docz` is running there.


**Enviroment**

- MacOS
- Tested with both Node 6.5.0 and Node 9.5.0

renatorib:
It's strange that it runs on 3001. Maybe the server is automatically identifying and using port + 1.

However you have several ways to use custom port. You can:

1. Add to you doczrs
```
export default {
  port: 4000,
}
```

2. Use cli arg
```
docz dev -p 4000
docz dev --port 4000
```

3. Use enviroment variable
```
PORT=400 docz dev
```

Any option should work.

Any doubt I am available 😄

raulfdm:
Hi @renatorib .

I had the same problem. I'm running a ruby server at `3000` and when I run `yarn docz dev` it said:
`You application is running at http://localhost:3000`.

I've checked about `http://localhost:3001` and it's running there, however, I think the message should tell us the correct used port! Might I open an issue to fix this message? Is this ticket can be used as a report? What do you prefer?

Bests.

pedronauck:
There's a bug here @raulfdm, we're using the default port to display the message, but when this port is changed, the message won't changes!

PR #107, opened 2018-06-30
Fix active port

good-idea:
### Description

Addresses #40 

This PR changes the location of the friendlyErrors `compilationSuccessInfo` message, allowing it to report the actual running port, instead of what was supplied in the config.

### Screenshots

| Before | After |
| ------ | ----- |
| image | image|

### DIFF
--- a/packages/docz-core/src/commands/dev.ts
+++ b/packages/docz-core/src/commands/dev.ts
@@ -1,6 +1,7 @@
 import * as fs from 'fs-extra'
 import logger from 'signale'
 import detectPort from 'detect-port'
+import friendlyErrors from 'friendly-errors-webpack-plugin'
 
 import * as paths from '../config/paths'
 import { Config } from './args'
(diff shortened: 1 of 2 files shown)


packages/docz-core/src/commands/dev.ts
```diff
@@ -21,7 +22,26 @@ export const dev = async (args: Config) => {
   const server = await bundler.createServer(bundler.getConfig())
   const app = await server.start()
 
-  app.on('listening', async ({ server }) => {
+  app.on('listening', async ({ server, options }) => {
```
pedronauck: I like to separate the concerns of each script, so I think that put bundler logic on `dev.ts` command is a mistake. Shouldn't be a function of a command update bundler's plugins.

And in this case, you can easily do something like we did to detect port of websockets:

```
const config = loadConfig(args)
const port = await detectPort(config.port)
const bundler = webpack({ ...config, port }, 'development')
```

pedronauck: Make this change you can keep doing plugin configuration on bundler instead of in the `dev.ts` command!

pedronauck (changes requested):

pedronauck:
Closed manually on commit 1b94114

PR #113, opened 2018-07-02
bug/ Make order ascending

### LINKED ISSUE #110: order not applying correctly
## Bug Report

**Describe the bug**

Using the `order` key in the `.mdx` file doesn't correctly reorder the menu items.

**To Reproduce**

1. Go to https://github.com/valtech-nyc/brookjs/pull/201
2. Clone the repository
3. Run `yarn && yarn docs:dev`
4. Open browser & verify menu order

**Expected behavior**

Menu should be reordered by `order` key.

**Enviroment**

* OS: OSX 10.13.5
* Node/npm version: Node 10.3.0/npm 6.1.0/yarn 1.7.0

brookjs

Swapnull:
### Description

Fixes #110. 
The order is currently descending, so when the order in files is 1, 2, 3 , it displays 

- 3
- 2
- 1

This feels backwards as I would expect to start an order from 1 and count up rather than start from 100 and count down. 
 
### Review

- [ ] Check the order

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,24 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+<a name="0.4.0"></a>
+# [0.4.0](https://github.com/pedronauck/docz/compare/v0.3.4...v0.4.0) (2018-06-30)
+
+
+### Bug Fixes
+
+* broken links to examples ([#101](https://github.com/pedronauck/docz/issues/101)) ([26458f2](https://github.com/pedronauck/docz/commit/26458f2))
+* **docz-core:** add a fallback name on Entry ([ec38139](https://github.com/pedronauck/docz/commit/ec38139))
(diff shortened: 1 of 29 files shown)


pedronauck:
Hmm, I think that's can be an unnecessary breaking change right now. Maybe in a possible release with a big breaking change, we can change this, but for now, I think that isn't the best option. I made the order ascending based on the `z-index` that I seem like a good option to put some things one above order.

What do you think guys? @Swapnull @marceloavf

pedronauck:
I thinking about that and I found a non-breaking change solution. Put some attribute on `doczrc` config to define is `order` is 'ascending' or 'descending' ✌️ Something like that:

```js
// docsrz.js
export default {
  order: 'ascending'
}
```

pedronauck:
Added manually on release [v0.5.0](https://github.com/pedronauck/docz/releases/tag/v0.5.0)

mAAdhaTTah:
The configuration gets me what I'm looking for, but it seems like a bit of unnecessary complexity. Maybe use it as a migration path?

mAAdhaTTah:
Right now, "descending" is default. If no value is provided, you can log a warning saying "In next release, we're going to remove this option; please change to "ascending" and update your project for forwards-compatibility." or something to that effect.

PR #116, opened 2018-07-04
corrected css class for PropsTable 'default'-column header

guustnieuwenhuis:
### Description

The 'default' column header in the PropsTable had a css class that's related to the 'description'-column. I've corrected this by giving the 'default'-column header a unique class.


nicholasess:
@Lagaffe you need to change commits for ```conventional comments```: https://conventionalcommits.org/

PR #119, opened 2018-07-04
feat(docz): use create-react-context

### LINKED ISSUE #118: Suggestion: Use create-react-context for 16.0+ compatibility
Currently, docz requires React 16.4+. I believe this is mostly due to the user of the new `createContext`, a feature landed in React 16.3 and used for docz's theming. For my project, I need to use React 16.2 because Enzyme, the testing library I'm using, doesn't yet support the new features added in 16.3, so I need to use React 16.2 in my project to run my tests.

If docz swapped out its usage of createContext for the [`create-react-context`](https://github.com/jamiebuilds/create-react-context) library, I'd be able to continue to use 16.2 in my tests.

I'd also be willing to do this work, if this is something you'd accept.

mAAdhaTTah:
Allows us to expand the allowable react version.

Fixes #118.


pedronauck (approved):

pedronauck:
Merged manually on 992eef9d

PR #140, opened 2018-07-13
feat(docz-core): expose filtered env variables

lupuszr:
### Description

Add dotenv to support reading .env files
Expose env varables to browser which start with
REACT_APP_
ANGULAR_APP_
VUE_APP_
NODE_

### Review

- [ ] Check the copy
- [ ] ...
- [ ] ...

### Pre-merge checklist

- [ ] ...
- [ ] ...

### Screenshots

| Before | After |
| ------ | ----- |
| Image  | Image |


packages/docz-core/src/bundlers/webpack/dotenv.ts
```diff
@@ -0,0 +1,88 @@
...
+const NODE_ENV = process.env.NODE_ENV
+// if (!NODE_ENV) {
+//   throw new Error(
+//     'The NODE_ENV environment variable is required but was not specified.'
+//   )
+// }
+
+// https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
+const dotenvFiles = [
+  `$dotenv}.${NODE_ENV}.local`,
```
pedronauck: Missing `{` here after `$`

packages/docz-core/src/bundlers/webpack/dotenv.ts
```diff
@@ -0,0 +1,88 @@
+import fs from 'fs'
```
pedronauck: please, move this file inside to `src/config` to be together with the `paths.ts` file...

pedronauck (changes requested):

packages/docz-core/src/bundlers/webpack/dotenv.ts
lupuszr: 👍

packages/docz-core/src/bundlers/webpack/dotenv.ts
lupuszr: 😱 Sure

PR #203, opened 2018-08-09
fix: allow code inside playground to contain ticks and backticks

### LINKED ISSUE #201: <Playground /> doesn't like '$'
## Bug Report

**Describe the bug**

Source:

```javascript
<Playground>
  Playground is fine with the $ character by itself
</Playground>
```

Result:

screen shot 2018-08-08 at 2 30 55 pm

---

Source:

```javascript
<Playground>
  Playground does not like the '$' character in single quotes
</Playground>
```

Result:

screen shot 2018-08-08 at 2 31 04 pm

**Expected behavior**

I expect that the playground result for the $ in single quotes will render properly, but it does not.

My guess is there's some regex magic going on and the source isn't being properly escaped.

rakannimer:
fixes #201

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "clean": "lerna clean",
     "bs": "lerna bootstrap",
     "packages": "run-s packages:*",
-    "packages:fix": "lerna run fix",
+    "packages:fix": "lerna run fix && echo",
     "packages:lint": "lerna run tslint",
     "packages:build": "lerna run build --ignore docz-example-*",
     "prerelease": "yarn run packages",
--- a/packages/rehype-docz/src/index.ts
+++ b/packages/rehype-docz/src/index.ts
@@ -28,11 +28,16 @@ const removePlayground = (code: string) => {
 const addCodeProp = async (node: any, idx: number) => {
   const name = componentName(node.value)
   const tagOpen = new RegExp(`^\\<${name}`)
-
   if (isPlayground(name)) {
     const formatted = await format(nodeToString(node))
     const code = formatted.slice(1, Infinity)
-    const child = strip(removePlayground(code)).trim()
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "clean": "lerna clean",
     "bs": "lerna bootstrap",
     "packages": "run-s packages:*",
-    "packages:fix": "lerna run fix",
+    "packages:fix": "lerna run fix && echo",
```
rakannimer: This is a hack, I had to do it because lint-staged was passing the file names as argument to `fix` script in packages, and it was breaking them, causing a false error.

kentcdodds: I have felt this pain before 😅

packages/rehype-docz/src/index.ts
```diff
@@ -28,11 +28,15 @@ const removePlayground = (code: string) => {
...
   const name = componentName(node.value)
   const tagOpen = new RegExp(`^\\<${name}`)
-
   if (isPlayground(name)) {
     const formatted = await format(nodeToString(node))
     const code = formatted.slice(1, Infinity)
-    const child = strip(removePlayground(code)).trim()
+    const child = strip(removePlayground(code))
+      .trim()
+      .split(`'`)
```
pedronauck: I think that is better to use `replace` instead of `join` and `split` in this case, no?!

pedronauck (changes requested):

pedronauck (approved):

packages/rehype-docz/src/index.ts
rakannimer: Yeah, I agree. Thanks for doing the change yourself #205 

Cheers !

PR #205, opened 2018-08-09
feat: allow edit code inside playground

pedronauck:


wachunga:
Is this documented anywhere? I was expecting to see an option in https://www.docz.site/documentation/components-api#playground

huchenme:
seems it is not working in `0.13`

chaseadamsio:
it does work if you wrap it in a closure:

```
  {() => {
    const someVar = 'value'
    return <MyComponent example={someVar} />
  }}
```

gives you an editable field for that code.

PR #255, opened 2018-08-25
fix(docz-core): join head tags

### LINKED ISSUE #248: htmlContext values aren't appended properly to <head> in cases of multiple of the same type.
## Bug Report

**Describe the bug**

When a htmlContext head value, such a links, have more than one value all but the first gets appended to the body with an accompanying "," element before each.

**To Reproduce**

Example of doczrc.js which reproduces the bug.

```
export default {
  htmlContext: {
    head: {
      links: [{
        rel: 'stylesheet',
        href: 'veryFake.css'
      },
      {
        rel: 'stylesheet',
        href: 'fake.css'
      }],
    }
  }
};
```

This produces html like this(all irrelevant html has been removed):

```
<head>
  <link rel="stylesheet" href="veryFake.css">
</head>
<body>
  ,<link rel="stylesheet" href="fake.css">
</body
```
**Expected behavior**

Both links should be in <head>

```
<head>
  <link rel="stylesheet" href="veryFake.css">
  <link rel="stylesheet" href="fake.css">
</head>
```

**Environment**

* OS: OSX 10.13.6
* Node: 10.9.0
* npm: 6.2.0
* yarn: 1.9.2

Screenshot of the html:
screen shot 2018-08-23 at 09 34 52

The culprit seems to be the parse-html util.
https://github.com/pedronauck/docz/blob/master/packages/docz-core/src/utils/parse-html.ts

The way the strings are being interpolated causes the issue.

```
"<div a>,<div b>" === `${['a', 'b'].map(i => `<div ${((item) => item)(i)}>`)}`
```

This is a runnable example which reproduces the logic inserting the head values into the headStr.
https://github.com/pedronauck/docz/blob/master/packages/docz-core/src/utils/parse-html.ts#L80
https://github.com/pedronauck/docz/blob/master/packages/docz-core/src/utils/parse-html.ts#L23

elv1n:
Fix #248 issue

### DIFF
--- a/packages/docz-core/src/utils/parse-html.ts
+++ b/packages/docz-core/src/utils/parse-html.ts
@@ -15,21 +15,21 @@ const wrapItems = (item: any) =>
     .map(key => `${key}="${item[key]}"`)
     .join(' ')
 
-const generateMetaTags = (items: any[] = []) => {
-  return items.map(item => `<meta ${wrapItems(item)}>`)
-}
+export type tagsTemplate = (type: string) => string
+const generateTags = (template: tagsTemplate) => (items: any[] = []) =>
+  items.map(template).join('')
 
-const generateLinkTags = (items: any[] = []) => {
-  return items.map(item => `<link ${wrapItems(item)}>`)
-}
+const generateMetaTags = generateTags(item => `<meta ${wrapItems(item)}>`)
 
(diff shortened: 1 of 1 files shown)


pedronauck (approved):

pedronauck:
thanks @elv1n 🙏

PR #263, opened 2018-08-29
refactor(docz-theme-default): use webpack to build instead of rollup

pedronauck:
This pull request change to use webpack to build docz-theme-default instead of rollup.


ematipico:
What's the reason that brought you to use webpack instead of rollup? Just curios

pedronauck:
`docz-theme-default` isn't a library or something that will be imported from another type of project except on docz. So, I don't need to bundle it with rollup, I can do this with webpack.

Using webpack instead of rollup the build that was 600k now is 70k.

PR #284, opened 2018-09-03
feat(docz): add support for exporting to CodeSandbox

CompuIves:
This allows you to export the code examples directly to CodeSandbox from the docs. 

We need to merge 

- https://github.com/Noviny/codesandboxer/pull/28
- https://github.com/Noviny/codesandboxer/pull/29
- https://github.com/mdx-js/mdx/pull/249

and update the dependencies first before we can merge.

screenshot 2018-09-03 23 46 18

screenshot 2018-09-03 23 54 20


jaburcodes:
@CompuIves Thanks <3 U're awesome

jaburcodes:
@CompuIves we were thinking about the development of react-native apps with Docz and doesn't make sense to show the CodeSandboxLogo when the developer is using the docz-plugin-react-native.

Instead, we were thinking of sending the user to https://snack.expo.io/

What's your opinion?

CompuIves:
> Instead, we were thinking of sending the code to snack.expo.io

Oh yes I agree! Exporting to Snack is perfect for react-native and mobile projects. I think we could do one of the two things for this PR:

1. Remove the button when we detect react-native-plugin.
2. Keep the button in for now, then the exported version will should work in CodeSandbox as we have support for `react-native-web`.

In the meantime we can look at what options Snack has for exporting. The current PR logic for exporting is pretty codesandbox focused (we use `codesandboxer-fs` for finding all the right file imports and generating the URL), but I can try to make it more generic so that Snack can also hook into the resulting files and dependencies. For me I prefer to do the export to Snack in a separate PR since I'm slightly afraid that this PR will take a long time to merge otherwise. But I'm fine with both directions, your call!

jaburcodes:
We had a discussion with the Docz core developers/community and on this PR we will send react and react-native-web codes to CodeSandbox as it already supports `react-native-web`(🙏).

Would be awesome if we could send the react-native-web code to Codesandbox with its preview screen resized to a phone default size, is it possible?

Later on, i think we will implement docz directly on the dev phone so the components can have access to native code and the phone API, we are discussing yet.

Thanks Ives.

CompuIves:
> Would be awesome if we could send the react-native-web code to Codesandbox with its preview screen resized to a phone default size, is it possible?

I like that suggestion, I can add a url prop called `dimensions` that will resize the preview to some dimensions. I will build that today.

CompuIves:
Hey! Almost everything is ready, the last thing I need to add is the detection of the type of component (React-Native vs React). Do you know how that currently works?

pedronauck:
I'll create some property on `doczrc` called `native` that is automatically set by `docz-plugin-react-native`, with this you can easy use this property to check @CompuIves :v:

pedronauck:
It's done in [v0.11.2](https://github.com/pedronauck/docz/releases/tag/v0.11.2), most specifically on [this commit](https://github.com/pedronauck/docz/commit/2580712) @CompuIves, now you can check `native` property on config, both on server and theme using `<ThemeConfig>`

CompuIves:
Updated! What's the best way to test React-Native?

pedronauck:
I think not @CompuIves, this last week was a mess for me, but I'll merge it today 🙏

TrySound:
Hi @CompuIves All examples here give 414 error. Is there a way to work around it?
https://istarkov.github.io/rifm/docs-readme

CompuIves:
Interesting, it seems like the generated `parameters` is not correct, I took it and tried to parse it here: https://codesandbox.io/s/208ooxkv8j. Do you know where I can find the source of the mdx? I'll do some debugging.

TrySound:
This one https://github.com/istarkov/rifm/blob/master/docs/readme.mdx
docz version is old there but I get the same after upgrade locally

TrySound:
Hi @CompuIves. Did you had a chance to debug this?

PR #291, opened 2018-09-05
feat(docz-core): generate static html based on route

mpivaa:
### Description

Make static building using routes as entry points
https://docz.canny.io/roadmap/p/static-build-based-on-generated-routes

### Pre-merge checklist

- [x] Generate multiple `index.html` based on components routes
- [x] Prerender components to improve SEO

### Bugs (https://dist-ggtvoutwro.now.sh/):

- [x] Emotion not emit style
- [ ] Playground layout

### Issues with prerender

Seems like there are some issues using `ReactDOMServer.renderToString` because of DOM dependencies. ~~I had some issues with dynamic importing the `root.tsx` component too~~, maybe there is a better way?


packages/docz-core/src/bundlers/webpack/config.ts
```diff
@@ -184,33 +189,50 @@ export const createConfig = (args: Args, env: Env) => async (
...
+          keepClosingSlash: true,
+          minifyJS: true,
+          minifyCSS: true,
+          minifyURLs: true,
+        })
+  }
+
+  if (args.entries) {
+    Object.values(args.entries).forEach((entry, idx) => {
+      config.plugin(`html-plugin-${idx}`).use(miniHtmlWebpack, [
```
mpivaa: `webpack-chain` doesn't seems to recognise multiple plugins with the same name or with multiple config objects in the array.

mpivaa:
@pedronauck @jaburcodes I'm having some problems with SSR because of codemirror dependency on DOM. Even if we fix that, knowing that the "theme" may be created by the user, this implicates in an uncontrolled environment that may lead to some problems in the future. Any thoughts?

pedronauck:
Do you talking about `document` and `window` issues @mpivaa? For sure, this can be a huge problem since each user can create your own theme. I think that we put all the benefits and problems of SSR and take a decision. If we follow the SSR path, we need to warn about these type of issues like NextJS does.

mpivaa:
@pedronauck by now, I've noticed a dependency of `window` on the `default-theme` and  `document` and `navigator` on `codemirror`, both solved using JSDOM and polluting node globals, which is bad.

Sure, I could run in a detached browser so the node globals remain intact, but I think I would need to make a partial bundle of the component (is there a way without this step?), which may over-complicate the build process.

Anyway, this doesn't help with the unpredictable user theme.

mpivaa:
I've just commited a prove of concept with SSR

mpivaa:
If we don't follow the SSR path, this would still be useful for SEO?

pedronauck:
@mpivaa good... I was thinking once to make something using Pupeteer, like get the markup of each static route using it and creating the generated HTML before the build process. This could give build slower but can be an alternative and don't need to use JSDOM. What do you think about? Makes sense for you?

mpivaa:
@pedronauck in terms of final result this is the best approach, but build time may be too slow because we would need to run webpack multiple times to transpile the code before it runs in pupeteer.

I was looking at x0 and it doesn't seem to care about this issues (even JSDOM is not being used)

mpivaa:
@pedronauck I've tested some approaches and the best result I think is the following:

- Use webpack to build bundle
- Start a http server pointing to the dist
- Use puppeteer to request and parse the routes
- Re-write static files with the result

One last issue: emotion is not extracting the CSS, already tested with `emotion-server` and it seems to work only when the component is being rendered in the same node process.

pedronauck:
Thanks @mpivaa, this is freaking awesome! We need to think how we'll resolve SSR issues around the CSS-in-JS, since each theme can define your own style system.

I'll check your branch and see this points 🤘

mpivaa:
Emotion issue fixed: https://dist-ggtvoutwro.now.sh/

geelen:
Just wondering if there's been any progress on this elsewhere? Looks like `docz` has changed a fair bit and there might be too many conflicts for this PR to be resurrected, but maybe the work can be restarted? I'm keen to help out if I can, I'm just sorta coming to it a bit late so I wanna know the best way to proceed. Cheers.

pedronauck:

PR #322, opened 2018-09-11
feat(docz): ordering menu on doczrc

mpivaa:
### Ordering menu on doczrc

Based on http://feedback.docz.site/roadmap/p/managing-menu-order-on-doczrc

### Pre-merge checklist

- [x] Order menu through doczrc
- [ ] Update docs

### DIFF
--- a/packages/docz/src/components/Docs.tsx
+++ b/packages/docz/src/components/Docs.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 import { Children } from 'react'
 import sort from 'array-sort'
 
-import { state, Entry, EntryMap, Config } from '../state'
+import { state, Entry, EntryMap, Config, MenuConfig } from '../state'
 import { entriesSelector } from './DocPreview'
 import { configSelector } from './ThemeConfig'
 
@@ -21,16 +21,19 @@ const comparePositionInConfig = (
   b: string,
   menu: string | null,
   config: Config
-) => {
+) => {  
 
   if(config.menu) {
-    const orderedMenuList = config.menu.map(m => m[0])
+    const orderedMenuList = config.menu.map(m => {
+      return typeof m === 'string' ? m : m.name
+    })
 
     if(menu) {
       const menuPos = findPos(menu, orderedMenuList)
 
-      if(menuPos !== UNKNOWN_POS) {
-        const orderedList = config.menu[menuPos][1] as (string[] | null)
+      if(menuPos !== UNKNOWN_POS && typeof config.menu[menuPos] === 'object') {
+        const menuConfig = config.menu[menuPos] as MenuConfig
+        const orderedList = menuConfig.docs
 
         return compare(
           findPos(a, orderedList),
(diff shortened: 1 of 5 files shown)


pedronauck:
Great job @mpivaa, you're doing a really great work, thanks 👏 

So, about this task, sorry to tell you just now, but I was thinking of something a little bit different and maybe a little bit simple too...

The idea is to use the `menu` property` to create the hierarchy and be able to add some custom properties for each one:

```js
export default {
  menu: [
    'Getting Started',
    {
      name: 'Components',
      menu: [
        'Alerts',
        'Button',
        ...
      ],
    },
    {
      name: 'Github',
      route: 'https://github.com/pedronauck/docz'
    }
  ]
}
```

But this is kinda a breaking change because of today when we use `<Docs>` component inside a theme to get the array of menus, this array just has strings. And to keep maintainability, we should need to use the `menu` to ordering and populate the data.

packages/docz/src/components/Docs.tsx
```diff
@@ -8,12 +8,60 @@ import { configSelector } from './ThemeConfig'
...
 export const isFn = (value: any): boolean => typeof value === 'function'
 
-const sortBy = (a: any, b: any, reverse?: boolean) => {
+const compare = (a: any, b: any, reverse?: boolean) => {
   if (a < b) return reverse ? 1 : -1
   if (a > b) return reverse ? -1 : 1
   return 0
 }
 
+const UNKNOWN_POS = 999999  // TODO: fix this?
```
pedronauck: I was thinking to try something like that:
https://github.com/mdx-js/mdx/blob/master/docs/_app.js#L49-L60

What do you think about @mpivaa ?

mpivaa: I thought about `Infinity` too, I will change

mpivaa:
I think these changes makes sense, but definitely they are breaking changes, maybe we should always pass an object as the `menu` and the theme handles how to present it?

pedronauck:
Yeah, for sure @mpivaa, this will be a breaking change definitely 😕 So, I think that is best we keep baby steps here and release this feature like people want, in a simple way and. Maybe leave this change to release in v1.0 and we can put some others nice things together ✌️

hnordt:
@pedronauck I can see you are not using `menu` in `doczrc.js`. Why you don't allow both passing `menu` and keep the current behavior when it's omitted. It would prevent a breaking change.

mpivaa:
@hnordt It would break if the user defines `menu` in `doczrc.js` but the theme is not prepared for an object.

mpivaa:
@pedronauck I've changed the config interface so it looks like:

```js
export default {
  menu: [
    'Getting Started',
    {
      name: 'Components',
      docs: [
        'Alerts',
        'Button',
        ...
      ],
    }
  ]
}
```

That way when we change the implementation of `Menu` to allow objects, it will not break old configurations.

pedronauck (approved):

alexlee-dev:
If I were to try and implement this now, would it work? v0.11.2

sidewalksalsa:
What version will this be released on @pedronauck @mpivaa ?

pedronauck:
Will be released in the `v0.12.0` @sidewalksalsa, I'm have just some bugs and performance improvements to solve before release this version 🙏

PR #324, opened 2018-09-11
[WIP] feat(docz): jsdoc integration

mpivaa:
### Description

JSDoc proof of concept integration


nicholasess:
@mpivaa Congratz, I was try with jsdoc-api, but parse it is way.

mpivaa:
@nicholasess I didn't know about `jsdoc-api` it has a function `explain` that do the same. In the end I think it is a better maintained package than `jsdoc3-parse`.

mpivaa:
@pedronauck my ideal solution was:

```
import { sum } from 'math'

<JSDoc of={sum} />
```

But we would need to make a babel plugin that injects the metadata during compilation.

mpivaa:
lodash `get`: https://lodash.com/docs/4.17.10#get
image

mpivaa:
Closing in favour of #375

Issue #351, opened 2018-09-21
Don't use unpkg/remote assets by default


kentcdodds:
I built a site and noticed it had this in the index.html:

`<link rel="stylesheet" href="https://unpkg.com/codemirror@5.39.2/lib/codemirror.css"/>`

I'd prefer that _all_ of the assets are local unless otherwise configured. I noticed there are other references to external CDNs in the styles as well. I'd prefer them all to be local so I can use the tool when I'm disconnected from the internet.

mpivaa:
I've just submitted a PR to fix this issue, let me know if there are more remote assets.

kentcdodds:
Perfect! Thanks!

IssuehuntBot:
@boostio funded this issue with $15. [Visit this issue on Issuehunt](https://issuehunt.io/repos/125596489/issues/351)

IssuehuntBot:
@mpivaa has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/125596489/issues/351)

IssuehuntBot:
@mpivaa has submitted [output](https://github.com/pedronauck/docz/pull/356). [Visit this issue on Issuehunt](https://issuehunt.io/repos/125596489/issues/351)

IssuehuntBot:
@pedronauck has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/125596489/issues/351)

PR #356, opened 2018-09-22
fix(docz-theme-default): use local vendors

mpivaa:
Fix: #351 

Styles are not being extracted by default so it's not required a new css import in the generated template, is this ok?


pedronauck:
I don't like so much the idea of put some vendors scripts hardcoded in the base, we'll need to change this scripts in each package update manually. Maybe add then as dependency and import can be a better way.

mpivaa:
Good point, fixed!

packages/docz-theme-default/webpack.config.js
```diff
@@ -40,10 +41,15 @@ const externalList = [
...
+const internals = [
+  'normalize.css',
+  'codemirror/lib/codemirror.css'
+]
+
 const deps = Object.keys(pkg.dependencies)
-const externals = Object.keys(pkg.dependencies)
+const externals = deps
   .concat(externalList)
-  .concat(deps.filter(dep => dep.startsWith('react-feather')))
```
mpivaa: Any reasons `react-feather` being add twice?

pedronauck (approved):

PR #380, opened 2018-10-01
feat(docz): ensure PropTable follow propTypes order

fkhadra:
### Description

The `PropTable` component doesn't follow the order defined by `propTypes`. I described everything under this [issue](https://github.com/pedronauck/docz/issues/379).

### How 

I used `Object.keys` on `component.propTypes` instead of the prop variable. I don't know if it's the best approach.

### DIFF
--- a/packages/docz/src/components/PropsTable.tsx
+++ b/packages/docz/src/components/PropsTable.tsx
@@ -129,7 +129,8 @@ const BasePropsTable: SFC<PropsTable> = ({ of: component, components }) => {
         </Thead>
         <Tbody>
           {props &&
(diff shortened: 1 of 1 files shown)


pedronauck:
Sorry dude, isn't possible accept your pull request, because this variable `props` came from react-docgen parsed props and this `component.propTypes` is just the prop types from `prop-types` package. If you just show properties from `component.propTypes` we will exclude props of flow and typescript parsed by react-docgen!

fkhadra:
Indeed, you're right. I forget about flow and typescript. I'll dig more thanks.

Issue #426, opened 2018-10-25
Use  `App` as component name in `Playground` will show nothing


qas612820704:
## Bug Report

**Describe the bug**

### It shows nothing

```jsx
---
name: Foo
---

import { Playground } from 'docz';
import App from './src/App';

<Playground>
  <App />
</Playground>
```

![screenshot from 2018-10-25 14-56-44](https://user-images.githubusercontent.com/8502252/47481558-7e2de180-d866-11e8-8785-a2134dd30525.png)

### It works fine

when changing `App` to `Bar`

```jsx
---
name: Foo
---

import { Playground } from 'docz';
import Bar from './src/App';

<Playground>
  <Bar />
</Playground>
```

![screenshot from 2018-10-25 14-57-34](https://user-images.githubusercontent.com/8502252/47481550-779f6a00-d866-11e8-82a1-88bfb2373c88.png)

### It also work fine when not wrap with `Playground`

```jsx
---
name: Foo
---

import { Playground } from 'docz';
import App from './src/App';

<App />

```

![screenshot from 2018-10-25 15-03-46](https://user-images.githubusercontent.com/8502252/47481752-38254d80-d867-11e8-8592-261de1646bbd.png)

**Expected behavior**

It should able to use `App` as component name in `Playground`

**Environment**

* OS: Linux T470 4.18.16-arch1-1-ARCH #1 SMP PREEMPT Sat Oct 20 22:06:45 UTC 2018 x86_64 GNU/Linux
* Node/npm version: [e.g. Node v10.12.0/npm 6.4.1]

PR #429, opened 2018-10-28
fix(docz-theme-default): rename playground container to avoid conflicts

beerose:
### Description

Solves issue: https://github.com/pedronauck/docz/issues/426
I renamed playground container to avoid conflicts with components named `App`.

### DIFF
--- a/packages/docz-theme-default/src/components/ui/Render/index.tsx
+++ b/packages/docz-theme-default/src/components/ui/Render/index.tsx
@@ -388,13 +388,13 @@ export class Render extends Component<RenderComponentProps, RenderState> {
 
   private transformCode(code: string): string {
     return `
-      const App = ({ children }) => (
+      const DoczApp = ({ children }) => (
         <React.Fragment>
           {children && typeof children === 'function' ? children() : children}
         </React.Fragment>
       )
 
-      render(<App>${code}</App>)
+      render(<DoczApp>${code}</DoczApp>)
     `
   }
 


packages/docz-theme-default/src/components/ui/Render/index.tsx
```diff
@@ -388,13 +388,13 @@ export class Render extends Component<RenderComponentProps, RenderState> {
 
   private transformCode(code: string): string {
     return `
-      const App = ({ children }) => (
+      const _App = ({ children }) => (
```
InquisitiveDevelopment:
```suggestion
      const _Docz_App = ({ children }) => (
```

What does everyone think about namespacing this so that it is even less likely to collide in the future? Since the user shouldn't need to interact with this directly I can't think of too many negatives to doing this.

packages/docz-theme-default/src/components/ui/Render/index.tsx
```diff
@@ -388,13 +388,13 @@ export class Render extends Component<RenderComponentProps, RenderState> {
...
         <React.Fragment>
           {children && typeof children === 'function' ? children() : children}
         </React.Fragment>
       )
 
-      render(<App>${code}</App>)
+      render(<_App>${code}</_App>)
```
InquisitiveDevelopment:
```suggestion
      render(<_Docz_App>${code}</_Docz_App>)
```

Jared-Dev (commented):
Nice changes! Thoughts on further namespacing that variable? While it is unlikely I could potentially see someone naming something with a single leading underscore.

packages/docz-theme-default/src/components/ui/Render/index.tsx
pedronauck: I think that use underscore is something weird, I prefer `DoczApp` like normal!

blackdahila: @pedronauck I renamed it do `DoczApp`

InquisitiveDevelopment:
Closes #426

Issue #485, opened 2018-12-04
Failed to mount: mount path must begin with "/"


kalynrobinson:
## Bug Report

**Describe the bug**

`yarn docz:dev` on a fresh create-react-app fails with the error:
> AssertionError [ERR_ASSERTION]: mount path must begin with "/"

**To Reproduce**

[Reproduction Repository](https://github.com/kalynrobinson/docz-reproduction)

(Optional: 1. Generate a fresh create-react-app and install Docz as usual)
2. Run `yarn docz:dev`
3. See error

```shell
create-react-app docz-bug
cd docz-bug
yarn add docz --dev 
# add scripts to package.json
yarn docz:dev
```

```shell
AssertionError [ERR_ASSERTION]: mount path must begin with "/"
    at mount (C:\Users\Kalyn Robinson\Documents\Development\aap\spill\node_modules\koa-mount\index.js:33:10)
    at Object.add (C:\Users\Kalyn Robinson\Documents\Development\aap\spill\node_modules\docz-core\dist\index.js:1:104906)
    at promise.then (C:\Users\Kalyn Robinson\Documents\Development\aap\spill\node_modules\docz-core\node_modules\webpack-serve\lib\app.js:45:17)
    at process._tickCallback (internal/process/next_tick.js:68:7)
error Command failed with exit code 1.
```

**Expected behavior**

Docz should successfully compile.

**Environment**

* OS: Windows 10
* Node/npm version: 
  * Yarn 1.12.3
  * Node v10.3.0

**Additional context/Screenshots**
Docz provides koa-mount with the path `\public`; koa-mount only accepts paths that begin with `/`.  Hardcoding the path as `/public` in `koa-mount/index.js` circumvents this issue (please forgive me for this sin).

pedronauck:
I tried this fresh setup using create-react-app and docz and everything worked fine, I think that this can be a windows problem 😕

kalynrobinson:
Thanks, I suspected as much! I'll have to take a crack at it this weekend. 😄

patricksevat:
Hi, I've run into this issue as well.

I've narrowed it down to https://github.com/pedronauck/docz/blob/master/packages/docz-core/src/webpack/devserver.ts#L46 . The fix is simple: just use `path.posix.join` to force a forward slash for the route as `path.join` will result in a backslash seperator on windows.

I can submit a PR if you'd like?

EduhCosta:
When the next release comes out, I'm in great need of this correction?

[...]

PR #491, opened 2018-12-06
fix(docz-core): #485 correct koa mounting path on windows systems

patricksevat:
### Description

See https://github.com/pedronauck/docz/issues/485 .

Examples appear unchanged. Might be good to check on a macOS machine as well.


Kivylius (approved):
Tested on windows 7. Working solution. :+1:

PR #530, opened 2018-12-20
Remove ownNodeModules & appNodeModules from config resolvers

jxom:
### Description

I'm not too sure if we need to specify `ownNodeModules` & `appNodeModules` when webpack should scan for `node_modules` by looking at the current directory as well as its relative parents (i.e. `./node_modules`, `../node_modules`, `../../node_modules`, etc). Was there any reasoning why it was done like this though? Correct me if I'm wrong!

I was having an issue where my app was depending on `package-a@1.0.0`, however, one of my app's dependants depended on `package-a@2.0.0`, however it was resolving to `package-a@1.0.0` (the app's `node_modules`).


pedronauck:
already included on v0.14.

jxom:
Awesome!

PR #535, opened 2018-12-20
fix: scroll to anchor if url has hash

### LINKED ISSUE #282: URL with anchor hash tag does not autoscroll upon first visit
## Bug Report

**Describe the bug**
When you go to a url like https://www.docz.site/documentation/creating-themes#default-theme-configuration, it will not go to the `default-theme-configuration` section as intended. Instead it just starts at the top. 

**To Reproduce**

1. Go to 'https://www.docz.site/documentation/creating-themes#default-theme-configuration'
2. See the auto scroll not happen

**Expected behavior**
It should go to the appropriate section

yaodingyd:
fix #282 
add a function in AsyncComponent to perform scrollIntoView in cDM

It doesn't work now because browser perform scroll to anchor when DOM is ready, and because of the async component, anchor would not be loaded into DOM yet. So I add a function in cDM lifecycle to check if we have hash in url, perform scroll to anchor.

### DIFF
--- a/packages/docz/src/components/AsyncComponent.tsx
+++ b/packages/docz/src/components/AsyncComponent.tsx
@@ -23,6 +23,7 @@ export class AsyncComponent extends Component<Props, State> {
 
   public componentDidMount(): void {
     this.fetch()
+    this.scrollToAnchor()
   }
 
   public render(): ReactNode {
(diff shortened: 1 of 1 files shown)


tibotiber:
Bump, any reaction to this?

pedronauck:
I got your commit by a cherry-pick since now both branches have a lot of conflicts.

bryanberger:
This behavior taken out of v1? implemented another way? @pedronauck

PR #542, opened 2018-12-21
fix: wrapper property is ignored in <Playground>

ddelbondio:
### Description

Fix wrapper property being ignored down the line the renderer of the default template.

I have changed the wrapping in the default theme renderer a bit to make sure that the wrapper is applied as close to the renderer playground as possible. This is specifically done to make sure the wrapper can reset all styles and there won't be any default styles defined inside the wrapper.

### BREAKING CHANGE
This changes the behaviour of \<Playground> in such way that instead of \<Playground> applying the wrapper, it is now the responsibility of the renderer to apply the wrapper.


ddelbondio:
darn, did not intend to make the PR against master. Sorry!

PR #571, opened 2019-01-09
change react-router to react-router-dom

angular-moon:
### Description

when `react-router-dom`  is treated as an external dependency in webpack config, an error will be caused:

` You should not use the < the Route > or withRouter () outside a < the Router > `

### Screenshots
image


pedronauck:
It's already merged using cherry-pick and included in the branch `v0.14`. Thanks!

PR #601, opened 2019-01-24
Move Prettier from dependencies to devDependencies in docz-utils

kachkaev:
Running `yarn why prettier` in my project, I noticed that Prettier is hoisted from `docz-plugin-css#docz-core#docz-utils#prettier`.  It should be moved to `devDependencies` if it's only used in scripts. Other docz packages might be worth checking too.

### DIFF
--- a/packages/docz-utils/package.json
+++ b/packages/docz-utils/package.json
@@ -40,7 +40,6 @@
     "jsx-ast-utils": "^2.0.1",
     "lodash.flatten": "^4.4.0",
     "lodash.get": "^4.4.2",
-    "prettier": "^1.15.3",
     "remark-frontmatter": "^1.3.1",
     "remark-parse": "^6.0.2",
     "remark-parse-yaml": "^0.0.1",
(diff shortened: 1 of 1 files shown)


kserjey:
It used as formatter for `<Playground/>` code:

https://github.com/pedronauck/docz/blob/daa1c89b5b033e323df06c18c40e7ac50aaad3bd/packages/docz-utils/src/format.ts#L1-L10

https://github.com/pedronauck/docz/blob/daa1c89b5b033e323df06c18c40e7ac50aaad3bd/packages/rehype-docz/src/index.ts#L22-L23

kachkaev:
I see. Is it a build-time dependency only or does Prettier also get into the client-side bundle? The module is compiled with rollup, which squashes all dependencies into rather massive monolithic JavaScript files. This can be a pretty heavy addition to the client side if there's not much value (the source mdx can get formatted with Prettier too by developers, so just trimming leading whitespace within the Playground can be enough).

If Prettier stays as a dependency, two minor bits to address:
* upgrade to 1.16.1
* [use `babel` instead of `babylon`](https://prettier.io/blog/2019/01/20/1.16.0.html#rename-babylon-parser-to-babel-5647-by-wuweiweiwu)

PR #609, opened 2019-01-29
fix: extend sidebar to the bottom of the page

woahitsraj:
### Description

Extends the sidebar on the default theme to the bottom of the page. Previously when scrolling down, the sidebar's background color would not display below the scroll point.

### Screenshots

| Before | After |
| ------ | ----- |
| screen shot 2019-01-25 at 11 12 30 | screen shot 2019-01-25 at 11 12 52 |


sw-yx:
looks good

PR #614, opened 2019-01-31
Fix/issue#528 Serve correct public folder

patricksevat:
### Description

Using `{ public: './public' }` in `doczrc.js` led to Express serving the wrong static path (at least on windows). This PR fixes that.

Tested on Windows 10 and on Ubuntu (on Windows Linux Subsystem) to test both win and posix file separators.

Tested with both relative paths and absolute paths on both OS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.13.7](https://github.com/pedronauck/docz/compare/v0.13.6...v0.13.7) (2018-12-26)
+
+**Note:** Version bump only for package docz
+
+
+
+
+
 ## [0.13.6](https://github.com/pedronauck/docz/compare/v0.13.5...v0.13.6) (2018-12-26)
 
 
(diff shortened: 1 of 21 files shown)


patricksevat:
I branched from master, let me know if I should rebase on dev

sw-yx:
where do we publish from? master or dev? i didnt realize there was a difference and now i'm really not sure

patricksevat:
Hi @pedronauck ,

Could you elaborate why you closed this PR? 
As far as I can test, this issue is still appears.

If you'd like I can reapply the fix to a new branch.

idmontie:
@pedronauck  @patricksevat Just commenting to mention that the wrong folder is served for static assets on my Mac and changing the line to `app.use('/public', express.static(publicDirPath))` fixes the issue.

pedronauck:
Sorry @patricksevat, I closed because your pull request was with a lot of diffs and this fix it's already done in the updated branch, I'm pretty hard working in a `v1` and it will be launched soon. So, this is kinda outdated.

patricksevat:
@pedronauck thx for the reply. Glad to hear you fixed the issue and I'm looking forward to v1!

PR #649, opened 2019-02-19
chore: upgrade prettier and tslint-config-prettier

kachkaev:
### Description

This PR upgrades dependencies related to Prettier and also [renames babylon parser to babel](https://prettier.io/blog/2019/01/20/1.16.0.html#rename-babylon-parser-to-babel-5647-by-wuweiweiwu) to avoid deprecation warnings. Running `yarn packages` locally went successful.


kachkaev:
A quick thought that came to my mind when preparing this PR. What about moving `fix:prettier` into the top-level `package.json`? It's a 'global' dependency rather than something applicable to individual components via `lerna`. With Prettier at the root of the project source, it will also cover files like `README.md`. Fixing the code with Prettier can be also [added as a pre-commit hook](https://prettier.io/docs/en/precommit.html).

I can submit a separate PR with this reshuffling if looks potentially useful.

PR #665, opened 2019-03-04
fix(gatsby-theme-docz): use options from gatsby-config

lettertwo:
**NOTE:** This is a PR against the v1 branch

### Description

The gatsby-node `onPreInit` hook passes in two arguments, the _second_ of which is the plugin options.

This fixes merging of options defined for gatsby-theme-docz in a project's `gatsby-config.js`.

### DIFF
--- a/core/gatsby-theme-docz/src/node/onPreInit.js
+++ b/core/gatsby-theme-docz/src/node/onPreInit.js
@@ -17,7 +17,7 @@ const touchTemplateWithPaths = paths => async (filepath, opts) => {
   await touch(dest, raw)
 }
 
-module.exports = async opts => {
+module.exports = async (_, opts) => {
   const { paths, ...config } = await parseConfig(opts)
   const componentPath = mountComponentPath(paths)
   const touchTemplate = touchTemplateWithPaths(paths)


core/gatsby-theme-docz/src/node/onPreInit.js
```diff
@@ -17,7 +17,7 @@ const touchTemplateWithPaths = paths => async (filepath, opts) => {
   await touch(dest, raw)
 }
 
-module.exports = async opts => {
+module.exports = async (_ opts) => {
```
zainxyz: Shouldn't this be `async (_, opts) => {` instead?

Missing the `,` (comma / period).

lettertwo: Yes, yes it should. Fixed!

pedronauck (approved):

pedronauck:
thanks @lettertwo, really... btw, did you testing v1?

lettertwo:
Yes! Using v1 alpha in a project.  I’ve got a few temporary hax in place to make it work, but it’s early days on the project, so I’m planning to keep up to date with the v1 progress.

Issue #552, opened 2018-12-24
TypeError: n is not a function


HanyRabah:
## Bug Report

*Something broken when webpack trying to compile with docz-core.*

**To Reproduce**
> Yarn docz dev

```shell
yarn run v1.12.3
$ docz dev 
✓ Creating app... 1 ms
✓ Parsing mdx files 74 ms
Happy[jsx]: Version: 5.0.0. Threads: 6 (shared pool)
➤ ｢wds｣: SockJS v0.3.19 bound to "/sockjs-node"
✓ Starting your server 1 ms
⠹ Initializing data server  77%                 234ms, estimated 303ms
Happy[jsx]: All set; signaling webpack to proceed.
✓ Initializing data server 273 ms
Happy[mdx]: All set; signaling webpack to proceed.
/Users/rabah/Projects/dbz-ui-framework/node_modules/docz/node_modules/docz-core/dist/index.js:1
(function (exports, require, module, __filename, __dirname) { "use strict";function _interopDefault(e){return e&&"object"==typeof e&&"default"in e?e.default:e}Object.defineProperty(exports,"__esModule",{value:!0});var detectPort=_interopDefault(require("detect-port")),equal=_interopDefault(require("fast-deep-equal")),remarkParse=_interopDefault(require("remark-parse")),glob=_interopDefault(require("fast-glob")),resolve=_interopDefault(require("resolve")),artTemplate=require("art-template"),doczUtils=require("docz-utils"),crypto=require("crypto"),slugify=_interopDefault(require("@sindresorhus/slugify")),pReduce=_interopDefault(require("p-reduce")),getPkgRepo=_interopDefault(require("get-pkg-repo")),findup=_interopDefault(require("find-up")),chokidar=_interopDefault(require("chokidar")),WS=_interopDefault(require("ws")),merge=_interopDefault(require("deepmerge")),getCacheIdentifier=_interopDefault(require("react-de

TypeError: n is not a function
    at parseHtml (/Users/rabah/Projects/dbz-ui-framework/node_modules/docz/node_modules/docz-core/dist/index.js:1:117684)
    at template (/Users/rabah/Projects/dbz-ui-framework/node_modules/docz/node_modules/docz-core/dist/index.js:1:119140)
    at MiniHtmlWebpackPlugin.plugin (/Users/rabah/Projects/dbz-ui-framework/node_modules/mini-html-webpack-plugin/index.js:17:33)
[...]

PR #688, opened 2019-03-18
refactor: tip for non-exist indexHtml option, as mentioned in #552

leftstick:
### Description

`TypeError: n is not a function` is not good for debugging the issue. Therefor, an explicit error with readable message would be better


pedronauck:

Issue #188, opened 2018-08-04
supply the readme in npm 


ycjcl868:
## Question
may supply the `readme` in the [npm](https://www.npmjs.com/package/docz)  , or looks like sham.

marwahaha:
I'm not sure if a symlink will work, but perhaps README.md in that package can point to the top-level file. You could try publishing and see what happens.

These 5 packages don't have a readme:
https://github.com/pedronauck/docz/tree/master/packages/docz
https://github.com/pedronauck/docz/tree/master/packages/docz-core
https://github.com/pedronauck/docz/tree/master/packages/load-cfg
https://github.com/pedronauck/docz/tree/master/packages/rehype-docz
https://github.com/pedronauck/docz/tree/master/packages/remark-docz

rakannimer:
Thanks for pointing this out 👍 I added READMEs that point to the main package with a small explanation of what the package does.

PR #712, opened 2019-03-24
chore: add README file to `docz` package

Glazy:
### Description

Resolves: #188

this was suggested in an issue and the rationale is that currently when
you view the `docz` package on NPM there is no description as to what it
is/what it does

### Review

- [ ] Check the copy


pedronauck:
There's some way to create a symbolic link instead of a copy and paste?

PR #615, opened 2019-02-01
chore: Export components from docz-theme-default

huchenme:
### Description

Export components from `docz-theme-default` so it will be possible to make a custom component with dynamic markdown content

I believe currently it is not possible to do something like this

```md
# Hello {firstName} {lastName}
```

But with this PR you could achieve

```jsx
import {components} from 'docz-theme-default'

<components.H1>Hello {firstName} {lastName}</components.H1>
```

and it will also possible to include markdown headers in a custom component and serve as some kind of layout component

PR #715, opened 2019-03-25
chore(docz-theme-default): export enhance from docz-theme-default

lttb:
### Description

In continuation of the discussion here #615 (comment) it would be great to have an ability to import default enhancer, that already configured for the docz components.

At this moment it's quite hard to use imported docz components, because they work with a specific context, that you should build by your side.


pedronauck (approved):

lttb:
thank you a lot!

Issue #736, opened 2019-03-29
Create component that displays component description rendered by react-docgen


jchadwick:
While the `<Props>` component is great, it does not include the component's description, if provided.   There should be another component which inspects the component's metadata like `Props` does, but displays the description instead of the props.

For example, given the following component:

```js
/**  The greatest button in the world */
class Button extends Component {
   /*...*/
}
```

and the MDX
```md
# Button
<Description of={Button} />
```

I would expect to see:
```
<h1>Button</h1>
<span>The greatest button in the world</span>
```

jesperorb-caspeco:
I did a quick component like that a few days ago. The description is located in `__docgenInfo.description`:
```jsx
import * as React from "react";

interface IDescriptionProps {
    of: React.ReactNode;
}

export default class Description extends React.PureComponent<IDescriptionProps, {}> {

    public getDocgenDescription = (comp: any) => {
        if(comp && comp.__docgenInfo) {
            return comp.__docgenInfo.description;
        }
        return "";
    } 

    public render() {
        return this.getDocgenDescription(this.props.of);
    }
}
```

jchadwick:
Closing issue due based on feedback.

PR #737, opened 2019-03-29
feat(docz): add Description component

jchadwick:
### Description

Adds the `<Description>` component to print component descriptions, as per #736 

### Example

*MyComponent.tsx*
```jsx
import React from "react";

/**
 * The greatest component in the world
 */
export const MyComponent = () => <div />;

export default MyComponent;
```

*MyComponent.mdx*
```
import { Description } from "docz";
import MyComponent from "./MyComponent";

# My Component

<Description of={MyComponent} />
```

### Screenshots

| Before | After |
| ------ | ----- |
|  image | image |

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,10 +99,12 @@ Next, add some `.mdx` files anywhere inside your project:
 name: Button
 ---
 
-import { Playground, Props } from 'docz'
+import { Description, Playground, Props } from 'docz'
 import Button from './'
 
 # Button
+
+<Description of={Button} />
 
 <Props of={Button} />
 
--- a/core/docz/src/components/Description.tsx
+++ b/core/docz/src/components/Description.tsx
@@ -1,0 +1,30 @@
+import { get, last } from "lodash/fp";
+import * as React from 'react';
+import { ComponentClass, createElement, FunctionComponent, SFC } from 'react';
+import { doczState } from '../state';
+import { ComponentWithDocGenInfo } from "./Props";
+
+type Tag = FunctionComponent | ComponentClass | string;
+
+export interface DescriptionProps  {
+  of: ComponentWithDocGenInfo;
+  as?: Tag;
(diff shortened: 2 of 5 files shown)


jchadwick:
@pedronauck what do you think about adding this?

pedronauck:
Hi Jess, sorry for the delayed answer. So, about this component, I think that is so minimal its behavior and I don't know it will be used at all, the component description can simply write on your mdx however you want. You can use it on your project, but to put inside the core I think that's not a good idea.

jchadwick:
Hey Pedro, thanks for responding!

With all due respect, to say that the component description can be written on your mdx misses the point that JSDoc documentation is just as useful for in-editor documentation as it is for generating static documentation.  Frankly, the in-editor documentation like shown below is the _more_ important use case.

image

In using docz without this component I have to provide the component description in the JSDoc and then (as you said) also copy and paste the same description into the MDX, which I've found to be very tedious, inefficient, and more difficult to maintain.  In other words, I think it's a bad practice.

I think just having this component readily available will encourage people to "do the right thing" and put descriptions on their components since it is then very easy to display those descriptions. I think it makes a very natural sibling to the `Props` component.  And, since it is so simple it should not be a headache to maintain in core.

cupofjoakim:
I have to agree with @pedronauck here. If you (@jchadwick) feel that this is the right way I'd urge you to maintain it as a separate repo. If it gains traction I think it'd be fair to think about merging it into this repo, but as a general consumer of docz I can't see the immediate added value.

jchadwick:
Withdrawing PR

Glazy:
@jchadwick If you maintain this as a separate package then it may be worth adding a section to the documentation to let users know about it. Whilst it has been decided this won't be added to the main repo right now I'm sure some users will find it useful :+1:

PR #751, opened 2019-04-01
fix: be sure to test the last version and prevent first run to fail

ivan-dalmet:
Getting an error when running `yarn packages` for the first time.
image

So I moved the `packages:test` after `packages:build` in the `package.json` to be sure that we test the last version and prevent first tests to fail.


Glazy (approved):
Looks good to me @ivan-dalmet :shipit:

pedronauck (approved):

PR #760, opened 2019-04-03
Props components not working with default exports

lesha1201:
## Issues
https://github.com/pedronauck/docz/issues/717
https://github.com/pedronauck/docz/issues/758

## Notes
Make sure you cleared cache before checking this. I had issues with babel cache (helped when I uninstall all `node_modules` and caches folders, and wrote `yarn`, `yarn bs`, `yarn packages:build`)


pedronauck:
@lesha1201 I updated the docgen typescript algorithm to have some kinda a cache, because it was so slow and this was a big problem for large bundles. Please, resolve the conflict, then I can merge your pull request :pray:

lesha1201:
@pedronauck Ok, I will do it tomorrow or day after tomorrow

lesha1201:
@pedronauck But there is a problem, that was fixed in docgen.ts - several exports works incorrectly

apvarun:
@pedronauck This Doesn't resolve the path correctly when the folder structure is as follows.

- src
  | index.jsx     --- this file imports and exports all components
  - Component
    | index.jsx      --- component file

I'm currently importing the Component via the parent index.jsx.


In this case, the metadata added should include the file name as `**src/Component/index.jsx**` But it actually has the value `**src/index.jsx**`.

_I could have directly imported the component but have styles loaded from parent index.jsx file._

PR #766, opened 2019-04-04
fix: docgen parsing default export

pedronauck:


lesha1201:
@pedronauck But there is a problem, that was fixed in docgen.ts - several exports works incorrectly

pedronauck:
Oh, can you help with that @lesha1201?

lesha1201:
@pedronauck I'm working on it and I could fix it in TS, but I found that docgen-typescript doesn't work properly with `const a..., b` and `export { a, b }`. I'm trying to figure out how to solve it

PR #771, opened 2019-04-06
Typescript doesn't work with several exports

lesha1201:
## Issues
https://github.com/pedronauck/docz/issues/769

## Notes
I found that `react-docgen-typescript` doesn't work with `consts` and several `export specifiers`. I opened an issue about this: https://github.com/styleguidist/react-docgen-typescript/issues/176


pedronauck:
Thanks again @lesha1201 🙏

PR #774, opened 2019-04-07
feat(docz-core): auto open browser when start dev mode

fi3ework:
### Description

Auto open bowser when running dev mode(just like create-react-app). Or we could determine if auto open browser in command, such as `docz dev -o`. For now, I made it as default cause auto-open directly might be more in line with the intuition.🤗

### DIFF
--- a/core/docz-core/src/commands/dev.ts
+++ b/core/docz-core/src/commands/dev.ts
@@ -41,7 +41,7 @@ export const dev = async (args: Arguments<any>) => {
 
   try {
     await dataServer.start()
-    openBrowser(`http://${config.host}:${config.port}`)
+    if (args.open || args.o) openBrowser(`http://${config.host}:${config.port}`)
   } catch (err) {
     logger.fatal('Failed to process data server')
     logger.error(err)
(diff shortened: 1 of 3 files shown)


Glazy:
Nice @fi3ework! :fire:

Not sure what @pedronauck thinks (?) but I think it would be good to introduce this functionality behind a flag like you suggested (`-o`) :+1:

pedronauck:
I think that since this feature is something native in `webpack-dev-server` we could have some argument like `-o`, but use can just use the flag and set the option here:
https://webpack.js.org/configuration/dev-server/#devserveropen

fi3ework:
hi:
I made it optional now, the argument to enable it is `docz dev --open` or `docz dev -o`. But for now still not use the native open browser method from webpackDevServer. The open file is copied from `create-react-app`. I guess this custom open function might be more robust and reliable than the native open browser method from webpackDevServer(not confirmed, just speculate). Any suggestions for improvement? 🤗

fi3ework:
I took a closer look at the code, may it's better to implement openBrowser in plugins, just adjacent to `friendlyErrors`

pedronauck (approved):

PR #779, opened 2019-04-09
filterComponents only works with capitalized files (Button.tsx, etc)

lesha1201:
### Description

filterComponents only works with capitalized files (Button.tsx, etc). But a lot of people use seperate folder and `index.*` for their components. If current behaviour is expected then I think it should be in documentation (I haven't found in Project configuration section)

P.S. it should be in documentation any way


pedronauck:
I think that this can be fixed by improving the regexp, I'll doo that :v:

PR #786, opened 2019-04-11
feat(docz-theme-default): add RunKit for code block

fi3ework:
### Description

Related roadmap: http://feedback.docz.site/roadmap/p/integration-with-runkit

Implement RunKit for code block. There is still some work to be done: 
- [x] dynamic inject the runkit vendor script. Now it's hard coded in html template.
- [x] better style for runkit text logo. I tried to import the `fira-sans` code but the minified html turn to be like `<linkhref="xxx" ..>`. May resolve it latter.

Any Suggestions for improvement? 🤗

### DIFF
--- a/core/docz-core/src/config/argv.ts
+++ b/core/docz-core/src/config/argv.ts
@@ -73,7 +73,6 @@ export interface Argv {
   websocketHost: string
   native: boolean
   codeSandbox: boolean
-  runKit: boolean
   sourcemaps: boolean
   /* template args */
   title: string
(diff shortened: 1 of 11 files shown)


fi3ework:
hi @pedronauck !
A few remaining bugs have just been fixed. Runkit can be switched in `doczrc.json`

```
export default {
  runKit: true,
  menu: ['Getting Started', 'Components'],
}
```
Works like this: 

Jietu20190414-133049-HD

pedronauck:
Hey @fi3ework, thanks a lot, I'll review this in the weekend since I need to think about, but in the first look I guess that this is freaking good :v:

PR #805, opened 2019-04-17
fix(babel-plugin-export-metadata): re-export causes meta error

### LINKED ISSUE #794: "export { default } from" not working
## Bug Report

**Describe the bug**

I have several index files that re-export the default value from another file:
```js
export { default } from "./Button";
```

All of these cause the same error on build:
```
Module parse failed: Unexpected token (3:11)
You may need an appropriate loader to handle this file type.
| export { default } from './Button';
|
> if (typeof default !== 'undefined' && default && default === Object(default)) {
|   Object.defineProperty(default, '__filemeta', {
|     enumerable: true,
```

Any other type of export seems to work, but would involve changing a lot of files.

**Environment**

* OS: OSX 10.14.4
* Node/npm version: Node v8.15.1, npm 6.4.1

fi3ework:
### Description
This PR could fix #794 . @pedronauck 
Fix re-export caused error. Following re-export are fixed: 

```js
export { default } from './a'

export { a } from './a'

export { default as b } from './a'

export { a as default } from './a'

export { a as b } from './a'

```

### Review

- [x] add tests

### DIFF
--- a/other-packages/babel-plugin-export-metadata/src/index.js
+++ b/other-packages/babel-plugin-export-metadata/src/index.js
@@ -15,6 +15,11 @@ const buildFileMeta = template(`
   }
 `)
 
+const replaceExportDefault = template(`
+  import NAME from 'SOURCE'
+  export default NAME
+`)
+
 const getFilename = state => {
   const filename = get(state, 'file.opts.filename')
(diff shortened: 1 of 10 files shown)


pedronauck:
Real cool dude, thanks a lot 🙏

PR #818, opened 2019-04-19
Reduce ci logging

BBB:
### Description

This patches the `0.13.7` release with the ability to reduce logs in a `process.env.CI` environment. (See: https://github.com/pedronauck/docz/issues/597).

Would it be possible to merge it into a `0.13-*` branch, and create a `v0.13.8` release? We're quite a way off being able to migrate to the `v1` release, and this is currently spitting out waaay too many logs.


kachkaev:
Same here. We're on 0.13 and can't upgrade to 1.0.0 because of https://github.com/pedronauck/docz/issues/763. Merging this into 0.13 and releasing 0.13.8 would be very useful!

rakannimer:
Hey @BBB 

Sorry for the late reply and thank you for taking the time to create this PR.

Unfortunately, we don't have, at the moment, a separate maintained branch for 0.x versions and the affected code doesn't exist in v2 so this PR can't be merged. 

Thanks again !

PR #819, opened 2019-04-19
feat(docz-core): add params to support dev server in the cloud

bcbrian:
### Description

I do a fair amount of developmnent in the cloud. The current version does not work in the cloud do to a localhost secuirity fix in webpack. I look at how create react app solved and issue and more or less borrowed what they did.


pedronauck:

Issue #825, opened 2019-04-22
Component Props not rendering


jsplopes:
The component props isn't rendering.

On prop component the **stateProps** key have "/" as path separator (unix slash):

![image](https://user-images.githubusercontent.com/35458912/56505599-f6bee380-6513-11e9-8592-3b6e5c23a48c.png)

And in the **filename** is used the "\" as path separator (windows backslash):

![image](https://user-images.githubusercontent.com/35458912/56505543-dabb4200-6513-11e9-94d6-c494936e495b.png)

The path validation must be agnostic, so maybe in the conditional found the path separator
on both must be normalized to a unix slash.

I'm using:
- Docz v.1.0.0
- Typescript
- dev compile on windows 10

regards,
Júlio Lopes

PS: Nice job...

fi3ework:
Could be fixed in next released version.

Glazy:
Closing, follow this issue in #777 :muscle:

PR #829, opened 2019-04-23
fix(babel-plugin-export-metadata): transform to Unix slash (#825)

fi3ework:
### Description

transform Windows backslash to Unix slash


other-packages/babel-plugin-export-metadata/src/index.js
```diff
@@ -22,7 +22,11 @@ const replaceExportDefault = template(`
 
 const getFilename = state => {
   const filename = get(state, 'file.opts.filename')
-  return filename && path.relative(process.cwd(), filename)
+  let filePath = filename && path.relative(process.cwd(), filename)
```
patricksevat: Probably `return filename && path.posix.relative(process.cwd(), filename)` would be enough. Unsure what this does exactly, but in previous versions of Docz often changing to path.posix was enough.

fi3ework: hi @patricksevat 
I'm not so familiar with `path` module. I searched what can `path.posix.relative` do and seems that it could not tranform '\\` to '/` in win32 platform (https://github.com/nodejs/node/issues/13887). And currently there is no `path.posix` code in docz. This pr's main target is tranform '..\\a\\b\\c' to '../a/b/c' in win32.

ejuo (approved):

Issue #783, opened 2019-04-10
Docz not building on Windows


ejuo:
## Bug Report
Get error when build Docz v1.0.0-rc.8 (all v1 same error).

```
...
...
...
src/ast.ts, src/codesandbox.ts, src/format.ts, src/fs.ts, src/index.ts, src/imports.ts, src/jsx.ts, src/mdast.ts → lib...
[!] Error: './format' is imported as an external by src\codesandbox.ts, but is already an existing non-external module id.
Error: './format' is imported as an external by src\codesandbox.ts, but is already an existing non-external module id.
    at error (C:\Git\docz.next\docz-master\node_modules\rollup\dist\rollup.js:3598:30)
    at C:\Git\docz.next\docz-master\node_modules\rollup\dist\rollup.js:18014:21
    at process._tickCallback (internal/process/next_tick.js:68:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:757:11)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
error Command failed with exit code 1.
```
to fix this I take only 1 entry 'src/index.ts' in [rollup.config.js](https://github.com/pedronauck/docz/blob/master/core/docz-utils/rollup.config.js).
Then after successfully build, all tests failed
```
Test Suites: 1 failed, 1 total
Tests:       6 failed, 6 total
Snapshots:   6 failed, 6 total
Time:        6.887s
Ran all test suites.
error Command failed with exit code 1.
```

**To Reproduce**

1. clone docz.
2. `yarn install`
3. `yarn bs`
4. `yarn packages`

I suspect, rollup don't resolves dependences, all bundles has links to src .ts files.
Example:
index.esm.js from [docz-utils](https://github.com/pedronauck/docz/tree/master/core/docz-utils) contains
```
export * from './ast.ts';
export * from './codesandbox.ts';
export * from './fs.ts';
export * from './imports.ts';
export * from './jsx.ts';
export * from './mdast.ts';
export { format } from './format.ts';
```
**Expected behavior**

It should build completly bundles.

**Environment**

* OS: Windows 10
* yarn 1.13.0 (or npm 6.5.0), 
* node 11.13.0 (or 10.15.0)

fi3ework:
[...]

PR #830, opened 2019-04-24
Docz external resolve fix

ejuo:
Fix for #783


pedronauck:
Can you explain to me how this could improve or what this change? Because all dependencies and external are already not bundled.

core/docz-theme-default/rollup.config.js
```diff
@@ -1,12 +1,5 @@
 import { config } from 'docz-rollup'
 
 export default config({
-  input: 'src/index.tsx',
-  external: id =>
-    !id.startsWith('\0') &&
-    !id.startsWith('.') &&
-    !id.startsWith(process.platform === 'win32' ? process.cwd() : '/') &&
-    !id.startsWith('@components') &&
```
pedronauck: And about this?

ejuo:
> Can you explain to me how this could improve or what this changes? Because all dependencies and external are already not bundled.

This is an alternative (cross-platform) solution for searching external dependencies. Do you think the expression like !id.startsWith ('.') is safe?

pedronauck:
I never have any problem with this, but we can keep both I think, we don't need to remove the default external configuration and the external configuration configured inside the modules, just add this plugin in the default plugins.

pedronauck (approved):

PR #845, opened 2019-04-30
Add missing LICENSE.md files

gregorym:
### Description

Large organizations usually do not use npm as a node package registry. Instead they take a snapshot of npm and exclude packages with security vulnerabilities or strict/missing licenses.

This PR is adding missing licenses to docz can be imported.


gregorym:
Hey,
When are you cutting a new release?

Issue #827, opened 2019-04-22
Docgen does not pick up types from d.ts declaration files


joeljeske:
## Bug Report

**Describe the bug**

The docgen props parser properly picks up types found in TS source files, but it does not pick up types that declared in d.ts files next to .js files.

I have some custom build settings that would be difficult to replicate in webpack and want to build my packages first, and then have Docz find the related MDX files that import the build .js and d.ts packages.

**Expected behavior**

I would expect the props parser to find the .d.ts files and find the related prop types to be included in the docgen output.

**Environment**

* OS: OSX 10.13.6
*  Node 8/npm 5

ejuo:
I also thought about it.
Originally planned to separate the library and documentation in different repositories.
But there is the big problem in that when building the library, the react components lack metadata and cannot be unambiguously compared with the source files (including d.ts).
But from a practical point of view, there is such an opportunity.  Do not strictly match the names of files and properties of components (eg. `displayName`), in most cases this will be enough. For this you need:
1. For each documented component, specify `displayName` property, which is equal to its file name.
2. Modify the doczrc.js like this:
```
export default {
  typescript: true,
  docgenConfig: {
	  propFilter: () => true,
	  searchPath: 'node_modules/your-component-library',
	  },
  filterComponents: (files) =>
    files.filter(filepath => /\/[A-Z]\w*(.d)?\.(js|jsx|ts|tsx)$/.test(filepath))
}
```
Also minor improvements of the `Docz` are required. [PR](https://github.com/pedronauck/docz/pull/846) I prepared.
@pedronauck, what do you think?

joeljeske:
Interesting. Yea it would be nice if there was better support. 

This was very important for me so i ended up with the following workaround: 


```js
//doczrc.js

export default {
  filterComponents: () => Object
    .entries(require('./package.json').dependencies)
    .map(([dep]) => {
        const { types } = require(`${dep}/package.json`);
        if (types) {
          return path.relative(__dirname,  require.resolve(`${dep}/${types}`));
        }
    })
    .filter((typesPath) => !!typesPath)
}
```

And then I re-wrote the paths for the db props inside a "custom" theme that just redirects into the default theme


```js
import Theme from 'docz-theme-default';
import { createElement, FC, useMemo } from 'react';

export const ThemeWrapper: FC<any> = ({
  db,
  ...props
}) => {
  // We have to re-write the props so we can support .d.ts files instead of JS files
  if (db && Array.isArray(db.props)) {
[...]

PR #848, opened 2019-05-02
feat(docz-core): add config docgen ts resolver

ejuo:
### Description
#827
extended name resolver configuration for docgen typescript.


pedronauck:
Hey @ejuo, I won't accept this pull request because I'm working to add the new version of `react-docgen` that has typescript support included.

PR #855, opened 2019-05-07
feat(docz-core): add serve command

arthurdenner:
### Description

Hi! :wave: This PR adds a new command, `docz serve`, which builds the site using `docz build` and then serves the output directory as a static site spawning [`serve`](https://github.com/zeit/serve).

PS: I don't know why, but the visual output is not the same as when you run `serve` from the command line. I'll try to figure it out later, but it doesn't prevent the merge if everything works as expected.

### DIFF
--- a/core/docz-core/src/commands/serve.ts
+++ b/core/docz-core/src/commands/serve.ts
@@ -1,0 +1,11 @@
+import spawn from 'cross-spawn'
+import { Arguments } from 'yargs'
+import { parseConfig } from '../config/docz'
+import * as paths from '../config/paths'
+
+export const serve = async (args: Arguments<any>) => {
+  const config = await parseConfig(args)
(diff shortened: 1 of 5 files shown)


core/docz-core/src/commands/serve.ts
```diff
@@ -0,0 +1,11 @@
+import spawn from 'cross-spawn'
+import { Arguments } from 'yargs'
+import { parseConfig } from '../config/docz'
+import * as paths from '../config/paths'
+
+export const serve = async (args: Arguments<any>) => {
+  const config = await parseConfig(args)
+  const dist = paths.getDist(config.dest)
```
arthurdenner: I believe that `cross-spawn` handles relative paths well (that means I could have used `config.dest` directly, but I decided to take a more secure approach.

pedronauck (approved):

Issue #850, opened 2019-05-02
HTML class "with-overlay" does not clear on menu select


alec-buda:
## Bug Report

The "with-overlay" class added to the HTML tag does not clear upon menu selection for mobile devices. 

This class prevents scrolling due to the overflow hidden style that it applies.

This bug has occurred in the past a couple times and seems to be back:
https://github.com/pedronauck/docz/issues/476
https://github.com/pedronauck/docz/issues/293

*A clear and concise description of what the bug is.*

**To Reproduce**

I have reproduced this issue in three different Gatsby environments. The easiest way to reproduce is: 

1. Navigate to the button page of the Gatsby starter located at:
https://gatsby-docz.netlify.com/docs/docs-button

2. Open up dev tools and change the viewing size to a mobile device.

3. Open up the mobile menu. You will notice the "with-overlay" class is applied to the html tag. The HTML tag will look as per below:
```
<html lang="en" data-react-helmet="lang" class="with-overlay">
```
4. Click on the "Getting Started" page. You will notice that the "with-overlay" class remains.

**Environment**

```
Windows 10 Pro Version 1803
gatsby --version
2.5.12
npm --version
6.4.1
node --version
v10.15.3
```
As well as testing on the starter I had originally seen this issue in my own project with the packages below:
```
{
  "name": "ultra-config-website-react",
  "description": "Ultra Config Website - Front End",
  "version": "1.0.0",
  "private": true,
  "author": "Alex Buda",
  "dependencies": {
    "@hot-loader/react-dom": "^16.8.6",
    "appversion": "^1.7.1",
    "brace": "^0.11.1",
    "docz": "^1.1.0",
    "docz-theme-default": "^1.1.0",
    "express": "^4.16.4",
    "gatsby": "^2.3.35",
    "gatsby-background-image": "^0.3.6",
    "gatsby-image": "^2.0.41",
    "gatsby-link": "^2.0.17",
    "gatsby-plugin-google-analytics": "^2.0.19",
    "gatsby-plugin-react-helmet": "^3.0.12",
    "gatsby-plugin-robots-txt": "^1.4.0",
    "gatsby-plugin-sass": "^2.0.11",
    "gatsby-plugin-sharp": "^2.0.35",
    "gatsby-plugin-sitemap": "^2.0.12",
    "gatsby-remark-images": "^3.0.11",
    "gatsby-remark-prismjs": "^3.2.8",
    "gatsby-source-filesystem": "^2.0.33",
[...]

Issue #476, opened 2018-11-28
why 'html' tag add a className 'with-overlay' ?


zWingz:
## Question

**Description**
![image](https://user-images.githubusercontent.com/13031838/49128190-325ad600-f305-11e8-96fe-cdbad88fc34c.png)

it will set `with-overlay` when `window.innerWidth < 1118`, but it will not remove when `window.innerWidth > 1118`

*Give more information about your doubt*

i don't know why set `hidden` in `html`

[basic demo](https://github.com/pedronauck/docz/tree/master/examples/basic) have same issus

nathunsmitty:
I'm also experiencing this issue with docz 0.12.13

pedronauck:
Fixed on [v0.12.14](https://github.com/pedronauck/docz/releases/tag/v0.12.14) ✅

Issue #293, opened 2018-09-05
Page inoperable when selecting pages from side menu.


johnpittman:
## Bug Report

**Describe the bug**

Unable to use the page after selecting a component from the sidebar.

*A clear and concise description of what the bug is.*

Clicking a nav link renders the page but there is some kind of modal of it that keeps the page from scrolling. The only way to get rid of it is to click the same nav link of the sidebar again. Re-clicking the same sidebar link toggle the modal. I think this has to do with the sidebar.

```javascript
---
name: QuoteProgress
menu: components/interfaces
---

import { Playground, PropsTable } from 'docz';
import QuoteProgress from './index';

# QuoteProgress

## Properties

<PropsTable of={QuoteProgress} />

## Basic usage

<Playground>
  <QuoteProgress stages={[{
      name: 'Pick Consignee',
      key: 'instantQuote,',
      active: false,
    }, {
      name: 'Quote Info',
      key: 'containerInfo,',
      active: false,
    }, {
      name: 'Container Info',
      key: 'kio',
      active: false,
    }]} 
  />
</Playground>
```

![localhost_5555_](https://user-images.githubusercontent.com/6298219/45114572-f869a180-b112-11e8-9505-47546b4ad5eb.png)

**To Reproduce**

2. Click on a page from the sidebar menu multiple times to see the modal toggle

```shell
FOR BUGS: Insert debug trace
```

*In order to allow us to quickly reproduce you issue please include a link to a reproduction repository.*

**Expected behavior**

*A clear and concise description of what you expected to happen..*

**Environment**

* OS: [e.g. OSX 10.13.4, Windows 10]
* Node/npm version: [e.g. Node 8/npm 5]

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

paesvitor:
I got the same issue

pedronauck:
Yes, this is fixed in the new version that will be released tonight or tomorrow, thanks guys 🙏

PR #880, opened 2019-05-17
fix(docz-theme-default): ensures the overlay is hidden when the sidebar is collapsed

alec-francis:
### Description

Closes issue #850. This bug was also seen in earlier issues (#476, #293). 

Perhaps if we merge this pull request then we can have the confidence that the overlay will always be removed when the sidebar is hidden.

### DIFF
--- a/core/docz-core/src/bundler/devserver.ts
+++ b/core/docz-core/src/bundler/devserver.ts
@@ -17,7 +17,7 @@ export const devServerConfig = (hooks: ServerHooks, args: Args) => {
     publicPath: '/',
     compress: true,
     logLevel: args.debug ? 'debug' : 'silent',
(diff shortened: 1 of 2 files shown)


pepibumur:
Is there any plan to merge this one?

lemon-31:
@pepibumur I hope so too.

@pedronauck, what are your thoughts about this one?

pedronauck:

madou:
sweet, thanks @pedronauck 

will anything be done about scroll to top on navigation atm? any recommendations for fixing this in user land if we're waiting for v2?

pedronauck:
I'll release v2 in the next few days, I think is best to wait until there. It's very close :v:

PR #886, opened 2019-05-21
Ensure Required only shows up when required

benbayard:
### Description

In typescript props can be both required and have a default value. This ends up looking like `required` and `default=` are stacked next to of each other. Instead, if the value is required _and_ there is a default value, don't show the prop!

#### For instance

```tsx
interface Props {
  isClickable: boolean
}
class MyComponent extends Component<Props> {
  static defaultProps = {isClickable: true};

  render() { return null }
}

ReactDOM.render(<MyComponent />) // no error! 
```

In this case the prop is listed as required in the interface, but it isn't required _in usage_ because internal to the react component it knows the value is going to exist! 

### Review

- [ ] Is this how we should handle this behavior?

### Screenshots

| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/core/docz-core/src/bundler/devserver.ts
+++ b/core/docz-core/src/bundler/devserver.ts
@@ -17,7 +17,7 @@ export const devServerConfig = (hooks: ServerHooks, args: Args) => {
     publicPath: '/',
     compress: true,
     logLevel: args.debug ? 'debug' : 'silent',
-    clientLogLevel: args.debug ? 'info' : 'none',
+    clientLogLevel: args.debug ? 'info' : 'silent',
     contentBase: [nonExistentDir],
     watchContentBase: true,
     hot: true,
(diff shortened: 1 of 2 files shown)


benbayard:
ping @pedronauck I'd like to get this in! Let me know what you think.

benbayard:
Hey @rakannimer do you know why this was closed? Is it no longer necessary? 
Thanks

rakannimer:
Hey @benbayard 

Sorry for closing without an explanation !  

Thanks for the contribution, however v2 doesn't have a `docz-theme-default` to edit. It was moved to a Gatsby theme [gatsby-theme-docz](https://github.com/doczjs/docz/tree/master/core/gatsby-theme-docz)

If you could do the changes there I'd be happy to merge them !

PR #887, opened 2019-05-22
fix(docz-core): fix `open` argument for `dev` command

daiyam:
### Description

This change change the default value of argument `open` for the command `dev` to match the declared option in `core/docz-core/src/config/argv.ts`

```js
.option('open', {
  alias: 'o',
  describe: 'auto open browser in dev mode',
  type: 'boolean',
  default: false,
})
```

The issue was that `docz dev` was opening the browser even with the option `--no-open`.


daiyam:
#891 is better!

PR #888, opened 2019-05-23
Set devServerConfig to respect args.open

maktouch:
### Description

This commit makes the`--open` flag work as it should. Without this patch, the browser always opens automatically


pedronauck:

PR #900, opened 2019-05-28
add filetype .styl

cangku:
### Description
Fix Error: Could not create Open in CodeSandbox Error.

### Screenshots
| Before | After |
| ------ | ----- |
| image | It's ok |


pedronauck:
This is no longer necessary since the [new v2](https://github.com/pedronauck/docz/pull/950)

PR #902, opened 2019-05-28
fix(docz): component Props false positives when looking in state

johnnorris:
### Description

When rendering `Props` we've found that the component can match the wrong file. E.g. rendering the Props from `elements/Button.jsx` might accidentally show the props for `elements/SubmitButton.jsx` as the match is only done on `${componentName}.`.

Propose including a leading slash at the start, which should ensure this can't happen.


pedronauck:
Merged manually into `feat/gatsby`.

PR #903, opened 2019-05-28
Update dependencies (react-live@2.1.2)

gregorym:
The existing react-live version has a dependency on a library that is not MIT.

### Description

More details on the react-live PR: https://github.com/FormidableLabs/react-live/pull/147


pedronauck:
This is no longer necessary since [new v2](https://github.com/pedronauck/docz/pull/950).

pedronauck:

Issue #904, opened 2019-05-30
Use types and interfaces inside Playground


nicolasbouffard:
## Question

**Description**

Say I have the following working MDX :
```
<Playground>
  {() => {
    const person = {
      name: "John Doe"
    };

    return <div>{person.name}</div>
  }}
</Playground>
```
If I want to use TypeScript to write a type definition for the Person, I can't. The simple fact of adding a type or interface yields a `SyntaxError: Unexpected token, expected ";"` message :
```
<Playground>
  {() => {
    type Name = string;

    interface Person {
      name: Name;
    }

    const person: Person = {
      name: "John Doe"
    };

    return <div>{person.name}</div>
  }}
</Playground>
```
I can't add return types to functions either (same error message) :
```
<Playground>
  {() => {
    // This does not work :
    // const getHelloWorldMessage = (): string => "Hello world !"

    // But this does :
    const getHelloWorldMessage = () => "Hello world !"

    return <div>{getHelloWorldMessage()}</div>
  }}
</Playground>
```
I can import typed elements from outside the MDX and it works fine though :
```
// utils.tsx
export const getHelloWorldMessage = (): string => "Hello world !"
```
```
import { getHelloWorldMessage } from "./utils"

<Playground>
  {() => {
    return <div>{getHelloWorldMessage()}</div>
  }}
</Playground>
```

How can we do that ?

Edit : Just found out that we can't use the `rest` operator when destructuring props inside the playground either :
```
<Playground>
  {() => {
    const HelloWorld = ({
      message,
      ...rest
    }) => {
      console.log(rest)
      return <div>{message}</div>
    };

    return <HelloWorld message="Hello world !" otherProp="Some other prop"/>
  }}
</Playground>
```
This yields `(LiveProvider, in Resizable (created by Playground)) TypeError: Cannot read property 'offsetWidth' of undefined`. Adding this to the previous use cases makes it very hard to use the Playground for anything a bit more complicated than very basic components (buttons, alerts, etc.). Am I missing something ? Pretty much all the examples I have found only do that, but I feel like Docz can be used to provide a clean documentation for so much more complex components, if it wasn't for these issues...

nicolasbouffard:
Since it doesn't seem possible for the time being to do what I want natively in the Playground, I had to use an alternative, which required some changes to the `docz-theme-default` and `docz` packages. I shared those modifications in the following PRs :

- Allow to override the playground's editor's code : #906 ;
- Allow to hide the LiveError overlay : #907.

[...]

PR #906, opened 2019-06-01
Allow to override the playground's editor's code

nicolasbouffard:
### Description

Currently the code that gets displayed in the playground's editor is the same as the one passed to the playground as its children. But because of issue #904, I had to be able to pass a simple component to the playground, while still showing the actual code of that component in the playground's editor.

This PR allows to pass an additional prop to the `Playground` component, called `editorCode`, which is the actual code (as a string) that will be shown in the editor, regardless of the code passed to the playground as its children.

For example :

- This caused issues :
```
<Playground>
  {() => {
    type Name = string;

    interface Person {
      name: Name;
    }

    const person: Person = {
      name: "John Doe"
    };

    return <div>{person.name}</div>
  }}
</Playground>
```
- This didn't :
```
// Examples.tsx
type Name = string;

interface Person {
  name: Name;
}

const person: Person = {
  name: "John Doe"
};

export const Example = () => <div>{person.name}</div>
```
```
import { Example } from "./Examples.tsx"

<Playground>
  <Example />
</Playground>
```
But doing so makes the playground's editor only show `<Example />`, which forces the user to open the `Examples.tsx` file to actually see the code used.

With this PR, I can do the following and have the expected result :
```
import { Example } from "./Examples.tsx"
const exampleCode =  "<The example's code (the actual way to get it can vary and is beyond the scope of this PR).>"

<Playground editorCode={exampleCode}>
  <Example />
</Playground>
```

Edit : This PR revealed an expected but undesirable effect of the playground for my use cases, which I described and fixed in #907.

### DIFF
--- a/core/docz-theme-default/src/components/ui/Playground/index.tsx
+++ b/core/docz-theme-default/src/components/ui/Playground/index.tsx
@@ -109,6 +109,7 @@ export const Playground: SFC<PlaygroundProps> = ({
   style,
   scope,
   wrapper: CustomWrapper = Fragment,
+  editorCode,
 }) => {
   const { themeConfig, native } = useConfig()
   const initialShowEditor = getter(themeConfig, 'showPlaygroundEditor')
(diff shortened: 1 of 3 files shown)


bryanberger:
Ran into the same problem trying to document our controlled components like Notifications, Popovers, Modals, etc. that require a button to be pressed to invoke.

eg:
```
  <Button type="primary" onClick={openNotification}>
    Open the notification box
  </Button>
```

PR #907, opened 2019-06-02
Allow to hide the LiveError overlay

nicolasbouffard:
### Description

Currently the playground's component preview **always** shows errors detected in the editor. But because of issue #904, I had to do PR #906. And because of that, the code in the editor, while perfectly valid, might be judged invalid by the playground, which will show the LiveError overlay on top of my perfectly rendered component.

This PR allows to configure the `docz-theme-default` to always hide the LiveError overlay, through a new `themeConfig` key called `showLiveError` (as boolean, default to true). For example :
```
// doczrc.js
export default {
  themeConfig: {
    showLiveError: false
  }
};
```

### To reproduce

Based on PR #906 :

- Example.tsx

```
import * as React from "react";

type Name = string;

interface Person {
  name: Name;
}

const person: Person = {
  name: "John Doe"
};

export const Example = () => <div>{person.name}</div>
```

- Example.mdx

```
---
name: Example
menu: Components
---

import { Playground, Props } from 'docz'
import { Example } from "./Example"
const exampleCode = `import * as React from "react";\n\ntype Name = string;\n\ninterface Person {\n  name: Name;\n}\n\nconst person: Person = {\n  name: "John Doe"\n};\n\nexport const Example = () => <div>{person.name}</div>\n`

# Example

<Playground editorCode={exampleCode}>
  <Example />
</Playground>
```

### Screenshots

| Before | After |
| ------ | ----- |
| before | after |

### DIFF
--- a/core/docz-theme-default/README.md
+++ b/core/docz-theme-default/README.md
@@ -71,6 +71,10 @@ const config = {
    */
   showPlaygroundEditor: true,
   /**
+   * Show/hide the LiveError overlay when errors are detected 
+   */
+  showLiveError: true,
+  /**
    * Set the numbers of max lines before scroll editor
    */
   linesToScrollEditor: 14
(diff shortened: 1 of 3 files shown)


pedronauck:
It's impossible to merge this pull request since the new v2. But thanks to the idea, it's already in the new version.

PR #924, opened 2019-06-18
fix(docz,docz-theme-default): remove react & react-dom from dependencies

### LINKED ISSUE #923: [docz | docz-theme-default] react & react-dom is in dependencies
## Bug Report

`docz-theme-default` and `docz`'s `package.json` has `react` & `react-dom` in both `dependencies` and `peerDependencies`. It makes all kind of unwanted behavior in monorepo setups. For example:

If you have `react` & `react-dom` installed in the parent directory (`./`) and `docz` installed on a child directory (`./site`) directory, it creates a nasty bug. If you use hooks in your application, this error pops up:

```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
```

And it is specifically caused by the point number 1 described in the error message. If you remove the parent directory's `node_modules` (i.e. `./node_modules`) everything works fine.

**To Reproduce**

1. `mkdir parent && cd parent && npm init`
2. `npm install --save-dev react react-dom`
3. `mkdir site && cd site && npm init`
4. `npm install --save-dev docz docz-theme-default`
5. This will install `react` & `react-dom` in `./site/node_modules` (which it should not).

**Expected behavior**

`react` & `react-dom` should only be in the `peerDependencies`

MunifTanjim:
### Description

fixes #923


draperunner:
https://github.com/pedronauck/docz/pull/882 aims to solve the same issue. I made a comment there to solve it like you did here, @MunifTanjim. But it seems @saimonmoore hasn't updated his PR yet.

@MunifTanjim I guess you also should remove react, react-dom from `gatsby-theme-docz`'s dependencies.

draperunner:
Also, you should update the installation instructions in README.md now that `react` and `react-dom` need to be installed as peers.

MunifTanjim:
> I guess you also should remove react, react-dom from gatsby-theme-docz's dependencies.

Already did.

> Also, you should update the installation instructions in README.md now that `react` and `react-dom` need to be installed as peers.

Done!

draperunner:
Hm, I can't see that you've changed anything for `gatsby-theme-docz`. Are you confusing it with `docz-theme-default`? `gatsby-theme-docz` doesn't even have any peerDependencies, so I think they need to be added there and removed from dependencies.

MunifTanjim:
> Hm, I can't see that you've changed anything for `gatsby-theme-docz`. Are you confusing it with `docz-theme-default`?

You are absolutely right 😲 So sorry about that.

> `gatsby-theme-docz` doesn't even have any peerDependencies, so I think they need to be added there and removed from dependencies.

Done! :sweat_smile:

draperunner (approved):
Great! :D

I hope @pedronauck will merge this. I think it will potentially close https://github.com/pedronauck/docz/issues/837, https://github.com/pedronauck/docz/issues/707 and https://github.com/pedronauck/docz/pull/882

pedronauck (approved):

pedronauck:
Merged manually into `feat/gatsby`.

PR #943, opened 2019-07-06
chore: updated mdx-js/react dependency

handfulglitter:
### Description

Update @mdx-js/react to newest version, will fix the issue:  https://github.com/pedronauck/docz/issues/932


pedronauck:
This is no longer necessary since the [new v2](https://github.com/pedronauck/docz/pull/950)

Issue #667, opened 2019-03-05
Proxy support


akharkhonov:
## Bug Report

I need to redirect part of my requests to some another url like in webpack by "proxy" option, but i think it's not working in docz.
**To Reproduce**
Apply this webpackOverlay config
```
const webpackOverlay = {
  devServer: {
    proxy: {
      '/api': {
        target: 'http://localhost:8080'
      }
    }
  }
};

module.exports = {
  modifyBabelRc: babelrc => Object.assign({}, babelrc, { plugins: [] }),
  modifyBundlerConfig: config => {
    return merge(config, webpackOverlay);
  },
  typescript: true,
  menu: ['Home', 'Components'],
};

```
and call fetch with some url which include 'api'

*A clear and concise description of what you expected to happen..*
I want to redirect specific requests to some another location
**Environment**

* OS: [e.g. OSX 10.13.4,]
* Node/npm version: Node 11.10, Yarn 1.13.0
* Docz: 0.13.7

Baoyx007:
+1

PR #948, opened 2019-07-10
 feat(docz-core): add proxy config

Baoyx007:
#667


pedronauck:

PR #1028, opened 2019-08-27
fix(docz-core): extend base config instead of overwriting user config

axe312ger:
repopened as requested with dev branch as root/base


rakannimer:
Thanks ! Will update here when it's published.

PR #1046, opened 2019-08-30
chore: Add node_modules to git ignore list

selbekk:
### Description

This pull request adds `node_modules` to our .gitignore file. This way, you don't accidentally end up including it in your git repo.

### Review

- [ ] Check that node_modules is spelled correctly


rakannimer:
Awesome, thanks for catching and fixing that 👍

PR #1063, opened 2019-09-02
feat(docz-core) parse both tsx? and jsx? files when using Typescript

wokakou:
### Description

when set `useTypescript : true` ,the **Props** dosen't render , which component end  with .js and .jsx

### Review

- [ ] Check the copy
- [ ] ...
- [ ] ...

### Pre-merge checklist

- [ ] ...
- [ ] ...

### Screenshots

| Before | After |
| ------ | ----- |
| Image  | Image |


rakannimer:
Thanks for the PR @wokakou !

PR #1073, opened 2019-09-04
feat: adds theme path override option

### LINKED ISSUE #997: Point to a specific subdirectory for theme overrides? (2.0)
## Question

**Description**
I'm looking for a way to point to a specific subdirectory for the gatsby theme overrides in 2.0

Currently, I have:
```
+-- doczrc.js
+-- .docz
+-- src
|   +-- gatsby-theme-docz
```

What'd I'm aiming for:
```
+-- doczrc.js
+-- .docz
+-- theme
|   +-- gatsby-theme-docz
```

I haven't come across a straight-forward way to accomplish this in the config. Has anyone had any luck?

Dangoo:
### Description
Resolves #997 

- [x] Have the custom path override default to src in the docz config
- [x] Add argument to ensureFiles with the path blob.
- [x] Update the watch function here to use path from config
- [x] Make sure to write them to the destination directory as src/gatsby-theme-* independently of the variable to make sure Gatsby interprets them correctly.

### DIFF
--- a/core/docz-core/__tests__/config.test.ts
+++ b/core/docz-core/__tests__/config.test.ts
@@ -1,7 +1,7 @@
 import * as path from 'path'
 import { setArgs } from '../src/config/argv'
 import { getBaseConfig } from '../src/config/docz'
-import { getThemeDir } from '../src/config/paths'
+import { getThemesDir } from '../src/config/paths'
 import yargs from 'yargs'
 
 describe('config.argv & config.docz', () => {
@@ -20,7 +20,7 @@ describe('getThemeDir', () => {
     //@ts-ignore
     const config = getBaseConfig(argv, {})
 
-    expect(getThemeDir(config)).toBe(path.join(config.root, '/src'))
+    expect(getThemesDir(config)).toBe(path.join(config.root, '/src'))
   })
 
   test('returns correct path for custom themesDir entry', () => {
@@ -28,6 +28,6 @@ describe('getThemeDir', () => {
     //@ts-ignore
     const config = getBaseConfig(argv, { themesDir: 'theme' })
 
-    expect(getThemeDir(config)).toBe(path.join(config.root, '/theme'))
+    expect(getThemesDir(config)).toBe(path.join(config.root, '/theme'))
   })
 })
--- a/core/docz-core/__tests__/config.test.ts
+++ b/core/docz-core/__tests__/config.test.ts
@@ -30,4 +30,12 @@ describe('getThemeDir', () => {
 
     expect(getThemesDir(config)).toBe(path.join(config.root, '/theme'))
   })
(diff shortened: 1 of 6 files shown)


Dangoo:
@rakannimer for the last bit I'm struggeling a bit with the contents of the different variables/paths within the theme watcher:
https://github.com/doczjs/docz/blob/d2aff82fb941ca4bcb1d2b01649655c84c94c5c8/core/docz-core/src/bundler/machine/services/watch-files.ts#L13

Here we need a check whether the `themeDir` differs from `'src'`. And if so, a rewrite of the filepath would be necessary.
Are these `filepath`s relative to the users project folder or to the docz package?

rakannimer:
The watcher emits absolute paths. 

We could do a simple replace first occurrence of `themeDir` with `src` but that could be a problem if `themeDir` is already in the absolute paths. 

For example if `themeDir = "theme"` and path is `/theme/a/docz-project/`

We'd need something to catch the last occurrence of themeDir before the first occurrence of `gatsby-theme-` 

Does that make sense ?

Dangoo:
> For example if `themeDir = "theme"` and path is `/theme/a/docz-project/`
> We'd need something to catch the last occurrence of themeDir before the first occurrence of `gatsby-theme-`

Makes total sense :)
My assumption would be that `themeDir = "theme"` must be located within `/theme/a/docz-project/` (`/theme/a/docz-project/theme/`) and `path.root` is always the users project path.

```js
// First create an absolute path for our `themeDir`:
const themeOverridePath = path.join(paths.root, args.themeDir);
// returns '/Users/.../theme/a/docz-project/theme'

// Second make the path to a given `filepath` relative:
const relFilePath = path.relative(filepath, themeOverridePath);
// returns '/gatsby-theme-docz/**/index.tsx'

// Prefix with `src`
const newRelFilePath = path.join('src', relFilePath);
// returns 'src/gatsby-theme-docz/**/index.tsx'

// Finally make absolute again
const newFilePath = path.resolve(paths.root, newRelFilePath);
// returns '/Users/.../theme/a/docz-project/src/gatsby-theme-docz/**/index.tsx'
```

Am I missing something?

rakannimer:
The logic looks solid to me.

Maybe we could also wrap the code you provided as is, in a function that takes in `rootPath` , `filepath` and `themeOverridePath` and returns destination path to make it more testable.

Dangoo:
@rakannimer PR is ready for review. Had to do a little refactoring since `filepath` was relative but it works fine now for default path and custom path incl. watching. Tested on a little test project and linking `docz-core` via `yarn link`.

rakannimer:
Thanks for the great work @Dangoo ! Will release it in a bit as `2.0.0-rc.34`

Issue #1100, opened 2019-09-11
when overriding theme, use background color for content


jstacoder:
## Bug Report

**Describe the bug**

when overriding the theme, it no longer uses the themes 'background color'

**To Reproduce**

create default site
override theme
set colors.background

**Expected behavior**

pages background color should be w/e was set in the theme

**Environment**

* OS: [e.g. OSX 10.13.4, Windows 10]
* Node/npm version: [e.g. Node 8/npm 5]

rakannimer:
Thanks @jstacoder for reporting and fixing this ! Released as 2.0.0-rc.40 👍

PR #1101, opened 2019-09-11
feat(gatsby-theme-docz): apply theme background color to content

jstacoder:
### Description

make sure to use theme background color 

#1100


rakannimer:
Hello ! 

Thanks for this PR.

I'm not sure I understand though, why are you setting the background property to `background` isn't that an invalid value ? 

Did you mean to use a variable for the background color ?

PR #1110, opened 2019-09-13
fix: remove additional editor styles

### LINKED ISSUE #1109: V2: Font-size of live editor too big
## Bug Report

The `font-size` of the Playgrounds live editor was set to `28px` some days ago - even though the the `LiveEditor` component has a style applied with `font-size: 18px;`
In my opinion `28px` is "slightly" too big…

https://github.com/doczjs/docz/blob/e500b81bec6e9b9bfac599e7b9ec424b4a71acde/core/docz-components/src/components/Playground/index.tsx#L95

fontSize

Dangoo:
### Description

Resolves #1109

### DIFF
--- a/core/docz-components/src/components/Playground/index.tsx
+++ b/core/docz-components/src/components/Playground/index.tsx
@@ -8,7 +8,6 @@ import {
   LiveEditor,
   LiveProviderProps,
 } from 'react-live';
-import { merge } from 'lodash';
 import { Resizable } from 're-resizable';
 import copy from 'copy-text-to-clipboard';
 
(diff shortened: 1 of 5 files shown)


rakannimer:
👍 

Do we also have to remove the empty `styles` array ? 

I might be mistaken but I remember that if for some reason the theme doesn't have a `style` property, it will throw with : `can't call reduce on undefined`.

Dangoo:
> Do we also have to remove the empty styles array ?
Not necessarily, just looked stale there. The typings around theming are not exactly precise and helpful here.
I'll have a deeper look.

rakannimer:
Thanks @Dangoo. 

Let me know when this is ready to merge, it looks good to me !

Dangoo:
> Let me know when this is ready to merge, it looks good to me !

Ready when you are! :)

rakannimer:
Released as `2.0.0-rc.41` : https://github.com/doczjs/docz/releases/tag/v2.0.0-rc.41

PR #1132, opened 2019-09-27
fix(gatsby-theme-docz): typecheck default props

### LINKED ISSUE #1130: Props breaks production build
## Bug Report

Static Website breaks when assigning a `boolean` value to an object key in `defaultProps` and `Props` component is used. This only happens on Production mode, dev mode works fine.

Screen Shot 2019-09-27 at 2 49 05 PM

**To Reproduce**

1. Create a functional component
2. Set a `defaultProp` with a `boolean` value
3. Build docz in production mode
4. Open the website, the page that renders the MDX file breaks

```shell
TypeError: n.defaultValue.value.replace is not a function
    at index.esm.js:1343
    at Array.map (<anonymous>)
    at is (index.esm.js:1339)
    at Zo (react-dom.production.min.js:150)
    at Mi (react-dom.production.min.js:178)
    at Ka (react-dom.production.min.js:232)
    at Ga (react-dom.production.min.js:233)
    at Ps (react-dom.production.min.js:249)
    at Ts (react-dom.production.min.js:248)
    at Ss (react-dom.production.min.js:245)
```

**Environment**

* OS: OSX Mojave 10.14.5
* Node/npm version: Node v10.13.0 / npm 6.4.1

jesperorb:
### Description

Fixes #1130 

Every parsed prop gets a defaultValue of the type `string` in dev. But it seems like the prod build converts the value from a string to the actual value `true` based on the screenshot in the issue, could this be true?

### DIFF
--- a/core/gatsby-theme-docz/src/components/Props/index.js
+++ b/core/gatsby-theme-docz/src/components/Props/index.js
@@ -5,12 +5,25 @@ import { jsx } from 'theme-ui'
 import { ChevronDown, ChevronUp } from '../Icons'
 import * as styles from './styles'
 
+const getDefaultValue = ({ defaultValue, type }) => {
+  if (!defaultValue.value || !type.name) return null
+  if (defaultValue.value === "''") {
+    return <em>[Empty string]</em>
+  }
+  if (type.name === 'string') {
+    return <em>{value.replace(/\'/g, '"')}</em>
+  }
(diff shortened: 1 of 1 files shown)


rakannimer:
This looks great. It's much better not to assume defaultValue is a string anyway 👍.

rakannimer:
Released as `2.0.0-rc.45`

PR #1136, opened 2019-09-29
fix(gatsby-theme-docz) check prop default value for flow types

jesperorb:
### Description

Default value check for proptypes also checks for flow types


rakannimer:
LGTM 👍

Issue #1042, opened 2019-08-29
"base" property does not work as expected


godmar:
When specifying a `base: /mybase/` in `doczrc.js`, and then running `yarn run docz build`, the created `.docz/dist` folder looks like this:
```$ ls .docz/dist/
404                                                                   component---docs-resources-papers-md-fe63ff69f05fbc4ade87.js.map
404.html                                                              component---docs-resources-urls-md-18ff15bc97f33d5c99ec.js
app-166636596e5034530b76.js                                           component---docs-resources-urls-md-18ff15bc97f33d5c99ec.js.map
app-166636596e5034530b76.js.map                                       component---src-pages-404-js-d2f6e42ec0b240a466f0.js
chunk-map.json                                                        component---src-pages-404-js-d2f6e42ec0b240a466f0.js.map
commons-7abd6c6896e587d23041.js                                       mybase
commons-7abd6c6896e587d23041.js.map                                   page-data
component---docs-index-mdx-eb57cbfd78a8e941cf11.js                    static
component---docs-index-mdx-eb57cbfd78a8e941cf11.js.map                styles.2543ff947a193d7f33b1.css
component---docs-lecture-1-lecture-1-mdx-e41e47345bd392371e15.js      styles-488c87c5398088946843.js
component---docs-lecture-1-lecture-1-mdx-e41e47345bd392371e15.js.map  styles-488c87c5398088946843.js.map
component---docs-lecture-2-lecture-2-mdx-0c06a42ea379446ff851.js      webpack-runtime-3b441f85f0b567d235ec.js
component---docs-lecture-2-lecture-2-mdx-0c06a42ea379446ff851.js.map  webpack-runtime-3b441f85f0b567d235ec.js.map
component---docs-resources-papers-md-fe63ff69f05fbc4ade87.js          webpack.stats.json
```

The file `.docz/dist/mybase/index.html` contains references such as: 
```
<link as="script" rel="preload" href="/component---docs-index-mdx-eb57cbfd78a8e941cf11.js"/>
```
that do not contain `mybase`.
There is no `index.html` file.

I was expecting to copy (or link) the .docz/dist folder to the server root directory of my Apache server and then be able to access it at `https://host/mybase`.  This does not work with these generated files.

I would appreciate guidance whether what I am trying to do is possible with docz: build and then copy to a subfolder hanging off the server's root, with all necessary files contained in that subfolder.

rakannimer:
[...]

PR #1137, opened 2019-09-29
fix(gatsby-theme-docz) default to text in code blocks if no language

jesperorb:
### Description

Defaults to `"text"` as language if no language is specified on code block as mentioned [here](https://github.com/doczjs/docz/issues/1042#issuecomment-531435802)


rakannimer:
Thanks 👍

Issue #986, opened 2019-07-31
Collapsible Menu


vrushank:
## Need Menu Group Collapsible

**Description**
*As per the demo docs I have seen all of them had menu collapsible. But in the new version, I guess I don't see the menu collapsible, It just creates a section instead.*

## EXAMPLE OF COLLAPSIBLE MENU

This is the menu from listed examples:
![Screenshot 2019-07-31 at 6 01 29 PM](https://user-images.githubusercontent.com/933954/62212053-94506b80-b3bd-11e9-8ff9-339de9963937.jpg)

## LATEST MENU

In the latest version the menu looks like below:

![Screenshot 2019-07-31 at 6 05 05 PM](https://user-images.githubusercontent.com/933954/62212189-dda0bb00-b3bd-11e9-9823-c577c1102502.jpg)

Am I missing something or is it removed?

Sorry, I'm new to the world of react, gatsby and docz.

Thanks a lot in advance.

Regards.

darrenhaken:
I am also seeing this, what's the fix?

ronaklalwaniii:
I am also seeing this, what's the fix? Any Update ???

jesperorb:
@darrenhaken @ronaklalwaniii 

This has not been added to the new theme.

Almost done with the functionality of this, some trouble with highlighting headings left. Adding this tomorrow or on tuesday.

jesperorb:
It's in the latest RC: `v2.0.0-rc.52`

PR #1141, opened 2019-09-30
fix(gatsby-theme-docz): add collapsible menu

jesperorb:
### Description

It went missing in v2, #986

This is how it would look efter these changes:

Kapture 2019-09-30 at 16 01 53


rakannimer:
Nice 👍

Issue #1014, opened 2019-08-21
v2: dependency was not found in monorepo


wokakou:
## Bug Report
**Describe the bug**

in root path, `yarn docz:dev ` fails to start with the errors:

- This dependency was not found:

the dependency is not installed in current package, but installed in the root path

**To Reproduce**

1. add .mdx file in one of  the package in a monorepo
2. yarn docz:dev in root path

**Environment**

* OS:  OSX 10.13.4,
* Node  version:  Node v10.15.3

**Additional context/Screenshots**

use create-react-app

lerna.json

```
{
  "packages": [
    "packages/sites/*",
    "packages/apps/*",
    "packages/base-libs/*",
    "packages/biz-libs/*",
    "packages/tools/*"
  ],
  "version": "independent",
  "useWorkspaces": true,
  "npmClient": "yarn",
  "command": {
    "publish": {
      "conventionalCommits": true,
      "message": "chore: publish"
    },
    "version": {
      "conventionalCommits": true,
      "message": "chore: version",
      "push": false
    }
  }
}
```

Issue #933, opened 2019-06-28
Docz is compiling entire project with webpack. when it should only compile files that are imported


hannadrehman:
## Bug Report

**Describe the bug**

I have a react monorepo project powered by lerna, where there are multiple packages build by single webpack config. 

one of the package is UI-Components which is a common repo for all re usable components.
and i am using `antd` as base Ui library. and i wanted to document these components.
all of these components are independent.
 
i have noticed that docz when started is compiling all the files and packages because of which build is failing and i get some useless errors. 

is there a way to configure docz to only compile files that are imported in mdx files. since it is using webpack it should follow the dependency graph simply following the imports. but in my case it is compiling files which are no were related to the component. and i am getting useles runtime errors. 

**Environment**

* OS: OSX 10.14.4
* Node/npm version: v10.0.0/ v5.6.0

zicodeng:
Can you provide your docz config? thanks

rakannimer:
You can use

`files` to supply an array of file glob patterns to compile
`src` to point to a specific to compile only
`ignore` to supply an array of file glob patterns to ignore

PR #1149, opened 2019-10-02
fix(docz-core): add support for hoisted dependencies

### LINKED ISSUE #1094: Module not found: Error: Can't resolve 'react' in monorepo.
## Bug Report

2.x releases of docz gives `Module not found` errors when trying to start it with `docz dev`.

**To Reproduce**

1. Clone [this repo](https://github.com/Mefu/docz-error-reproduce)
2. `yarn` to install dependencies.
3. `yarn workspace test-dev-1 start:docz` to reproduce error.
4. `yarn workspace test-dev-2 start:docz` to see it is working correctly with docz 1.x

**Expected behavior**

I believe, I did not miss anything in migration guide and this should work correctly.

**Environment**

* OS: Arch Linux, kernel 5.2.11
* Node/npm version:
   * yarn: 1.17.3
   * node: 12.9.1

* Project lives in a monorepo, possibly related.
* mdx files import components from files compiled by rollup, possibly unrelated.

rakannimer:
### Description

Before this PR, to use docz in a monorepo you would have to install and use it from the root of the repo only.

This PR allows docz to be used in packages with hoisted node_modules.

- Adds `repoRootPath` config variable to doczrc (defaults to `../`)
- Adds example of monorepo usage without `repoRootPath`
```sh
npx create-docz-app monorepo-docz-in-root --example monorepo
```
- Adds example of monorepo docz usage in package : 
```sh
npx create-docz-app monorepo-docz-in-package --example monorepo-package
```

Fixes #1094 #1014 #933


selbekk:
😍

I have the following structure in my monorepo (a design system):

```
- packages
  - package-a
  - package-b
  - package-c
- docs
```

I keep it outside of packages, since it's not really a package I'm releasing, but a docs website. It is, however, specified as a workspace in Lerna workspaces.

Will this change affect me at all?

examples/monorepo-package/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "private": true,
+  "workspaces": [
+    "packages/*"
+  ],
+  "name": "docz-gatsby-monorepo",
+  "version": "2.0.0-rc.41",
```
selbekk: This should probably get a version bump?

examples/monorepo-package/packages/basic/doczrc.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  menu: ['Getting Started', 'Components'],
+  repoRootPath: '../..',
```
selbekk: could we skip this by looking up the tree for a `package.json` instead? If you have something super special, I guess you could set this as a last resort - but stuff that just works is a great deal!

tmp.code-workspace
```diff
@@ -0,0 +1,23 @@
...
+    },
+    {
+      "path": "examples/basic"
+    }
+  ],
+  "settings": {
+    "typescript.tsdk": "node_modules/typescript/lib",
+    "javascript.format.enable": false
+  }
+}
```
selbekk: Should this be added, or ignored?

selbekk (approved):

tmp.code-workspace
rakannimer: Thanks for catching that !

examples/monorepo-package/packages/basic/doczrc.js
rakannimer: Yeah I don't like it either and was really hesitant to add (yet another) config variable, but I couldn't think of a way to find the monorepo root.

> could we skip this by looking up the tree for a package.json instead?

We could, but how would we decide which parent directory to use ? 

For example for the following structure :

```
monorepo/
	package.json
	packages/
		basic/
			package.json
			.docz/
				package.json

```
The `basic` package could have its dependencies in basic/node_modules or monorepo/node_modules depending on the workspace configuration.

Any thoughts on a way around this ?

examples/monorepo-package/package.json
rakannimer: This is the example package, it will never be published to npm, so the version doesn't really matter.

rakannimer:
> I have the following structure in my monorepo (a design system):
> 
> ```
> - packages
>   - package-a
>   - package-b
>   - package-c
> - docs
> ```
> 
> I keep it outside of packages, since it's not really a package I'm releasing, but a docs website. It is, however, specified as a workspace in Lerna workspaces.
> 
> Will this change affect me at all?

Inside docs you're probably using a gatsby app with the gatsby-theme-docz. 

With this change you should be able to use `docz` inside docs like you would in a normal repo by using mdx files only.

You could also use docz from the root of your workspace: 

1 - yarn add docz@next -W
2 - Scatter your mdx files inside package-a, package-b and package-c
3 - yarn docz dev|build|serve

examples/monorepo-package/package.json
selbekk: ah it's the version, not the dependency. My bad 🙈

examples/monorepo-package/packages/basic/doczrc.js
selbekk: 🤔
We could check for the existence of the gatsby binary file - that's the only use we have for this (at least for now, right)? For the project root, we could just find the package.json furthest up the tree.

selbekk:
👏👏👏

examples/monorepo-package/packages/basic/doczrc.js
rakannimer: It now looks for the first gatsby dependency up the tree and uses that .
I left `repoRootPath` in the config as an escape hatch, for people that have very particular setups.

selbekk: Amazing job 💪 remember to document it!

Issue #1091, opened 2019-09-09
Attempting to use the Props builtin component. It simply just never works. 


robjac:
## Bug Report

**Describe the bug**

Attempting to use the Props builtin component. It simple just never works. 
There are no errors.

**To Reproduce**
import Props from docz ( `import { Playground, Props } from "docz";`

**Expected behavior**

Should render a `<Props of={MyComponent}/>` table.

Right now it just never renders; and there are no errors.

**Environment**

* OS: OSX 10.14.6 (18G87)
* Node: 12.4.0

rakannimer:
Hello

Could you please provide : 

1. The code for `MyComponent`
2. The `docz` version you're using

So we can try and reproduce your problem ?

crusoexia:
+1

* docz version: ^2.0.0-rc.38
* mdx:

```mdx
---
name: Button
---

import { Props, Playground } from 'docz';
import { Button } from './Button';

## Props

<Props of={Button}></Props>

## Example

<Playground>
  <Button text="click me" />
</Playground>
```

* Button.tsx

```typescript
import React from 'react';

interface Props {
  text: string;
}

export const Button: React.FC = ({ text }: Props) => {
  return (
    <button>{text}</button>
  )
}
```

results:

![Screen Shot 2019-09-10 at 1 11 04 PM](https://user-images.githubusercontent.com/5928754/64585742-7e769300-d3cc-11e9-93d0-3871ac6b3ef3.png)

The project is just initailized by *create-react-app* with docz installed.

rakannimer:
Could you please try : 

```diff
- export const Button: React.FC = ({ text }: Props) => {
+ export const Button: React.FC<Props> = ({ text }) => {
```

Also, make sure you have `typescript` set to true in your `doczrc.js` file

crusoexia:
@rakannimer same result.

* Button.tsx

```typescript
import React from 'react';

interface Props {
  text: string;
}

export const Button: React.FC<Props> = ({ text }: Props) => {
  return (
    <button>{text}</button>
  )
}
```

* doczrc.js

```javascript
export default {
  typescript: true
}
```

rakannimer:
One last thing, 

Can you try adding the extensions when importing Button in your mdx file ? 

```js

import { Button } from "./Button.tsx"

```

crusoexia:
@rakannimer It works!

[...]

PR #1151, opened 2019-10-02
(docz-core) handle props on windows

jesperorb:
### Description

Draft for fix for finding and parsing files on Windows (#1091 )

Not tested more than the supplied test that I am getting back props when setting state. Primarily [fastglob needs forward slashed](https://github.com/mrmlnc/fast-glob#how-to-write-patterns-on-windows) which it was not getting.

* Is there occasions where forward slashes should not be used in the state of the application, when is it a problem?

### DIFF
--- a/core/docz-core/__tests__/props.test.ts
+++ b/core/docz-core/__tests__/props.test.ts
@@ -1,47 +1,86 @@
-import { getBaseConfig } from '../src/config/docz'
-import * as paths from '../src/config/paths'
 import * as fs from 'fs-extra'
-import { setArgs } from '../src/config/argv'
 import { getPattern, initial } from '../src/states/props'
-import { Params } from '../src/lib/DataServer'
-import yargs from 'yargs'
-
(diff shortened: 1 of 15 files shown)


core/docz-core/__tests__/props.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      expect(cache).toBeNull()
+      const config = getTestConfig()
+      const pattern = getPattern(config)
+      const data = initial(config, pattern)
+      const params = mockedParams()
+      await data(params)
+      expect(readCacheFile()).not.toBeNull()
+    })
+  })
+})
```
rakannimer: 👍

rakannimer:
Nice PR ( and tests ) ! 👍 

> Is there occasions where forward slashes should not be used in the state of the application, when is it a problem?

Edge case for sure but maybe when escaping a character ?

For example : 

```sh
mkdir -p test/a@b
cd test/a\@b
```

Issue #959, opened 2019-07-19
v2: yarn as undocumented dependency


kelkes:
## Bug Report

`npm run docz:dev` fails due to missing `yarn` installation.

**To Reproduce**

1. Dev maschine without yarn
2. run `npm run docz:dev`
3. See error

![docz-yarn](https://user-images.githubusercontent.com/204445/61516978-70c51280-aa06-11e9-9927-a55577be040f.png)

**Expected behavior**

docz should run either with npm or yarn.

**Environment**

* Ubuntu 19.04
* node 12.4.0
* npm 6.10.1

**Additional context/Screenshots**
I think its here: https://github.com/pedronauck/docz/blob/master/core/docz-core/src/commands/serve.ts

Is this something that will probably change in upcoming rcs? I am willing to help here.

callmemagnus:
Yes it would great to mention that in the documentation.

OR

Even better, not rely on `yarn` for that (?). We are not using yarn at work. I know I could "just install it" but I think it would be better with rely on it.

Reklino:
Yeah, same issue here. Would be great not to rely on yarn for this.

pedronauck:
Hi guys, sorry the too late response, I was on vacation so I couldn't see any issues until now. I'll try to figure out this and fix in order to launch in the stable v2 🙏

MountainDrew:
I think this would be great if this didn't rely on yarn. Is this be something I could help get going? I was looking through the codebase and a lot of the tasks do rely on yarn, e.g. linting. Is the plan to just move the external running scripts away from yarn?

MountainDrew:
I have gotten this far: https://github.com/pedronauck/docz/commit/c971e633d37f78396ca766a0691e11152084048e

But I am looking on how to reference the prebuild of this locally via npm. I have tried the following from another repo: `"docz": "file:../docz/core/docz",`  but I am not sure it is correct.

jesperorb:
> I have gotten this far: [c971e63](https://github.com/doczjs/docz/commit/c971e633d37f78396ca766a0691e11152084048e)
> 
> But I am looking on how to reference the prebuild of this locally via npm. I have tried the following from another repo: `"docz": "file:../docz/core/docz",` but I am not sure it is correct. If anyone could give me a hand with the local building that would be great.

Hi, I can try to help. But I am not sure what you mean by prebuilding locally? What's the scenario exactly?

MountainDrew:
[...]

PR #1153, opened 2019-10-03
Remove npm as dependency [wip]

amprew:
### Description

Working on https://github.com/doczjs/docz/issues/959.

This is currently in the stage of working with installed dependencies using yarn. But when installing via npm it still falls short. Maybe need some pre-hook for installing

### Review

- [ ] Check the copy
- [ ] ...
- [ ] ...

### Pre-merge checklist

- [ ] ...
- [ ] ...

### Screenshots

| Before | After |
| ------ | ----- |
| Image  | Image |

### DIFF
--- a/core/docz/package.json
+++ b/core/docz/package.json
@@ -16,9 +16,9 @@
     "README.md"
   ],
   "scripts": {
-    "dev": "cross-env NODE_ENV=development yarn build -w",
-    "build": "cross-env NODE_ENV=production rollup -c",
-    "fix": "yarn lint --fix",
+    "dev": "cross-env NODE_ENV=development npm run build -- -w",
+    "build": "cross-env NODE_ENV=production npx rollup -c",
+    "fix": "npm run lint -- --fix",
     "lint": "eslint . --ext .ts,.tsx",
     "precommit": "lint-staged"
   },
--- a/core/docz-core/.lintstagedrc
+++ b/core/docz-core/.lintstagedrc
@@ -1,3 +1,3 @@
 {
-  "*.{ts,tsx}": ["yarn fix", "git add"]
+  "*.{ts,tsx}": ["npm run fix", "git add"]
 }
--- a/core/docz-core/nodemon.json
+++ b/core/docz-core/nodemon.json
@@ -1,6 +1,6 @@
 {
   "ignore": ["dist/"],
   "//verbose": true,
-  "exec": "yarn build && node bin/index.js init",
+  "exec": "npm run build && node bin/index.js init",
   "ext": "js,ts,json"
 }
--- a/core/docz-core/package.json
+++ b/core/docz-core/package.json
@@ -12,13 +12,13 @@
     "README.md"
   ],
   "scripts": {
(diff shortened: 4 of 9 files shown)


core/docz/package.json
```diff
@@ -28,7 +28,7 @@
     "array-sort": "^1.0.0",
     "capitalize": "^2.0.0",
     "docz-components": "2.0.0-rc.48",
-    "docz-core": "2.0.0-rc.51",
+    "docz-core": "file:../docz-core",
```
MountainDrew: this will not stay. It is a means of testing

jesperorb: Could you change this to it only references the latest RC of `docz-core` and push. We could see if the CI tests pass or fail on it.

MountainDrew: This passes now in circle ci.

My only worry now, since yesterday, is in master the exec command has been moved to the bin node module path of gatsby. This will likely have an effect with this branch trying to find where the bin file is:
https://github.com/doczjs/docz/blob/1e3454fd77caff8678c812bf4fa80cbdc4f62e44/core/docz-core/src/bundler/machine/services/exec-dev-command.ts#L37

MountainDrew: I will try to find time later today to look at this after I rebase it on top.

jesperorb: @rakannimer has been working on that part. Maybe he knows what could go wrong.

rakannimer: > My only worry now, since yesterday, is in master the exec command has been moved to the bin node module path of gatsby. This will likely have an effect with this branch trying to find where the bin file is:

I'm sorry I'm not sure I understand, why would your branch cause issues with the change you highlighted ?

MountainDrew: Basically when I rebase I get this exception as the dependencies are not pulled through via yarn, but instead via npm. Is this a way the modules are resolved into the root directory?

image

(Just a note, when testing, I reference a local version of docz in my package.json which is actually in the core folder like so `file:path-to/docz/core/docz`) Am I making a mistake here? I didnt think so as it seems to work when installing via yarn

jesperorb: I think the problem is like this, sorry for not replying better and earlier. We do need to support only having `npm` installed when consuming the packages. That means that we should let user use `npm` when running `docz`. The best solution would be to detect if yarn is installed/used and the use yarn, if not use `npm`. This means that we should let yarn be optional when running as a user: `docz build` and `docz dev`

But internally we still want to use `yarn` to be able to utilize workspaces when building the packages and using lerna. Your PR as it is now changes all occurances of `yarn` to `npm` which also changes how the packages are being resolved and built internally.

MountainDrew: One way I suppose we could check is the `$npm_execpath`? How do you think we can handle the installing of the packages via npm since workspaces won't work without yarn?

rakannimer:
Thanks @MountainDrew for putting in the time to work on this PR. 

As @jesperorb noted, changing the `docz` dev environment to work without `yarn` is not currently possible or desirable.

In #1220 I replaced all external facing `yarn` usage to `npm` (`serve` and `build` commands). Released in `2.0.0-rc.68`.

PR #1168, opened 2019-10-08
fix(docz-core): index.tsx imports in index.mdx

### LINKED ISSUE #1164: Crash when importing `index.tsx` from `index.mdx`
## Bug Report

**Describe the bug**

When importing a component from a `index.tsx` file in a `index.mdx` at the same location, it crashes with:

> Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

**To Reproduce**

1. `$ npx create-docz-app docz-app-typescript --example typescript`
2. Create a random exported component in the file `src/index.tsx` (either named export or default export)
2. Change `src/index.mdx` to

```mdx
import Component from './'
// or
import { Component } from './'

<Component />
```

4. Run `npm run docz dev`
5. See error in browser

**Expected behavior**

It should render the component without any error.

PS: it works fine with javascript files, the error occurs for TS only.

**Environment**

- docz version: v2.0.0-rc.53
- OS: OSX 10.14.6
- Node/npm version: Node v10 / npm v6

Screenshot 2019-10-07 at 14 53 49

Kerumen:
### Description

Attempt to fix https://github.com/doczjs/docz/issues/1164

It seems that if we put `gatsby-plugin-mdx` (set in `gatsby-theme-docz`) AFTER `gatsby-plugin-typescript`, it fixes the problem.

I've looked at another Gatsby starter with TypeScript and MDX and they [do the same](https://github.com/tylergreulich/gatsby-typescript-mdx-prismjs-starter/blob/64c6f8a3fc7896b8cdf5319ca9441c3c536354a4/gatsby-config.js#L11).

I don't know the side-effects but I've tested the typescript example and it works fine.

### DIFF
--- a/core/docz-core/templates/gatsby-config.tpl.js
+++ b/core/docz-core/templates/gatsby-config.tpl.js
@@ -16,17 +16,18 @@ const config = {
     description: "<%- config.description %>"
   },
   plugins: [
-    {
-      resolve: 'gatsby-theme-docz',
-      options: <%- opts %>
-    },<% if (config.typescript) {%>
+    <% if (config.typescript) {%>
     {
       resolve: 'gatsby-plugin-typescript',
       options: {
         isTSX: true,
         allExtensions: true
       }
-    },<%}%><% if (isDoczRepo) {%>
+    },<%}%>
+    {
+      resolve: 'gatsby-theme-docz',
+      options: <%- opts %>
+    },<% if (isDoczRepo) {%>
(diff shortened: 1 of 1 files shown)


jesperorb:
@Kerumen Thanks for opening up a PR. :) I will check out the branch and test it as well.

PR #1173, opened 2019-10-14
chore(babel-plugin-export-metadata): don't add `__filemeta` if it already exists

Kerumen:
### Description

I'm currently moving my documentation (written with Docz) outside of my package.
This leads to different problems, one being the props not well resolved by the `Props` component due to `__filemeta.filename` mismatch.

To avoid this problem, I try to build my library with `babel-plugin-export-metadata` to add the proper `__filemeta.filename` key and not a "random" `index.es.js` built.

However, Docz rewrites every imports with `babel-plugin-export-metadata` [here](https://github.com/doczjs/docz/blob/cfe07b0c00fa5ff73bd194fbc48240a5315bc10e/core/gatsby-theme-docz/lib/onCreateBabelConfig.js#L10-L15). My `__filemeta` is hence overwritten. This adds a condition if the key already exists to skip it.

The key name being very specific, I guess we are safe to assume that if it exists, the component author know what he is doing.

### DIFF
--- a/other-packages/babel-plugin-export-metadata/src/index.js
+++ b/other-packages/babel-plugin-export-metadata/src/index.js
@@ -3,7 +3,7 @@ const { default: template } = require('@babel/template')
 const { get } = require('lodash')
 
 const buildFileMeta = template(`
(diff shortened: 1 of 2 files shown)


rakannimer:
👍 Thanks !

PR #1181, opened 2019-10-16
Pass the entry property down to the wrapper

### LINKED ISSUE #1177: Customize meta description
### Need

I'd like to leverage the shadowing of the [wrapper](https://github.com/doczjs/docz/blob/master/core/gatsby-theme-docz/src/wrapper.js) component to add a meta description to the generated pages using the description that is included in the page frontmatter:

```md
---
description: just a description
---
Page content
```
### Proposal
Support passing the page frontmatter to the wrapper component so that developers can use frontmatter's metadata from the component.

pepicrft:
Resolves https://github.com/doczjs/docz/issues/1177

### Description

Passes the entry property down to the `Wrapper` component, which will allow developers to use it for things like setting meta headers in the header.

### DIFF
--- a/core/gatsby-theme-docz/src/base/Layout.js
+++ b/core/gatsby-theme-docz/src/base/Layout.js
@@ -17,7 +17,7 @@ const Route = ({ children, entry, ...defaultProps }) => {
   if (!entry) return <NotFound />
   return (
     <MDXProvider components={components}>
-      <Wrapper>
+      <Wrapper doc={entry}>
         <Layout {...props}>{children}</Layout>
       </Wrapper>
     </MDXProvider>
--- a/core/gatsby-theme-docz/src/wrapper.js
+++ b/core/gatsby-theme-docz/src/wrapper.js
@@ -1,6 +1,6 @@
 import * as React from 'react'
 
-const Wrapper = ({ children, doc }) => (
+const Wrapper = ({ children }) => (
   <React.Fragment>{children}</React.Fragment>
 )
 export default Wrapper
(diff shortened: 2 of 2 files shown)


pepibumur:
@rakannimer let me know if there's anything else that I should do as part of this PR. I tried to run the tests locally but I got the following error:

```
docz-core:   ● Test suite failed to run
docz-core:     Cannot find module 'docz-core' from '/Users/pedropinera/src/github.com/doczjs/docz/core/docz-core/src/config'
docz-core:     However, Jest was able to find:
docz-core:      '../config/paths.ts'
```

pepibumur:
[Found the answer](https://github.com/doczjs/docz/blob/master/CONTRIBUTING.md#what-does-yarn-dev-do-) myself 😬

pepibumur:
@rakannimer eslint fails because the property is not being used. Do you think I should use `<Helmet/>` to populate the header metadata if the props is given or rather disable `eslint` for that line and let the users override how they want to use the prop?

selbekk:
@pepibumur you don't have to destructure out the `doc` property from `props` - it will still be available for when you shadow the component.

pepibumur:
> @pepibumur you don't have to destructure out the doc property from props - it will still be available for when you shadow the component.

Just pushed a commit following your suggestion.

selbekk (approved):

rakannimer:
Looks good @pepibumur thanks !

rakannimer:
Published to npm as `docz@2.0.0-rc.58`

PR #1189, opened 2019-10-22
chore(docz-core): fix `resolveApp` when called from the docz project

Kerumen:
### Description

The `paths.ts` file was used at first from the base project hence the root was, let's say:
```
/Users/foo/bar/project
```
Then, this file is also used inside the Docz project, this time the root is:
```
/Users/foo/bar/project/.docz
```

However, `resolveApp` wants to always resolve the files in the base project.
I added a ternary to go up from one folder when we are in the Docz project.

This fixes [here](https://github.com/doczjs/docz/blob/cfe07b0c00fa5ff73bd194fbc48240a5315bc10e/core/docz-core/src/utils/repo-info.ts#L10-L17) where we want the project's package.json and not the docz' generated one.

I also cleaned some unused variables and added the `Config` type.

### DIFF
--- a/core/docz-core/src/config/paths.ts
+++ b/core/docz-core/src/config/paths.ts
@@ -2,6 +2,8 @@ import * as fs from 'fs'
 import * as path from 'path'
 import * as resolve from 'resolve'
 
+import { Config } from './argv'
+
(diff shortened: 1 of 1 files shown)


rakannimer:
Looks great 👍

PR #1190, opened 2019-10-22
style(docz-components): fix live preview background

Kerumen:
### Description

I noticed a minor UI glitch with the live preview background and the radius. I moved the CSS property from the inner wrapper to the outer one.

### Screenshots

| Before | After |
| ------ | ----- |
| Screenshot 2019-10-22 at 14 02 58 | Screenshot 2019-10-22 at 14 28 02 |


rakannimer:
Thanks for fixing that ! I couldn't stop seeing it once I noticed it 😄

PR #1193, opened 2019-10-23
feat(docz): forward of prop to custom Props component

selbekk:
### Description

This pull request forwards the `of` prop from the `<Props />` used in MDX to any custom `Props` component passed to the `componentsMap` component.

I want to be able to use the component name in my Prop component to add some more information, so therefore I need access to the passed component.


rakannimer:
Makes sense ! Thanks 👍

Issue #1105, opened 2019-09-12
Shadow theme components and styles missing in production build when themesDir is provided


mrfelton:
## Bug Report

Seems like the issue in https://github.com/doczjs/docz/issues/1056 may not have been fully resolved. Still some things are not being set up properly in a production build.

It's ok if you run `yarn dev` before you build and serve, but if you don't then some things are missing from the production build, such as shadowed theme components and styles.

**To Reproduce**

Compare:
```sh
git clone https://github.com/LN-Zap/zap-tutorials
cd zap-tutorials
yarn develop
```
with:
```sh
git clone https://github.com/LN-Zap/zap-tutorials
cd zap-tutorials
yarn build
yarn serve
```

Notice that the header looks completely different - this is because the shadow components and styles are missing if you don't `dev` before you `build`.

rubencustodio:
I am also facing this issue and I won't be able to build a production version until this is fixed. My shadow components are based on my company's style guide and it won't look correct without it.

rakannimer:
Hey,

As @brandonarbini said in #967 the issue is that docz is not copying the base directory (defaults to `src`) to the `.docz` directory before building and relying on dev being called before build.

If someone would like to submit a PR to fix it I'd be happy to help/review/merge.

If you can't, no problem. I'll get to it later this week !

rakannimer:
Hey this was fixed in [2.0.0-rc.42]
(https://github.com/doczjs/docz/releases/tag/v2.0.0-rc.42)

Could you give the latest rc a try ?

mrfelton:
Hi,

No this still doesn't work correctly.

Compare a build of https://github.com/LN-Zap/zap-tutorials with/without doing `yarn dev` first.

rakannimer:
Should be fixed in 2.0.0-rc.51 tried zap-tutorials locally with it and it worked. Could you confirm ?

rakannimer:
Hmmm I think the update didn't go right, here's what it looks like after running yarn add docz@next on my installation :  

<img width="357" alt="Screen Shot 2019-09-30 at 5 00 26 PM" src="https://user-images.githubusercontent.com/1270909/65916733-dee16a80-e3a3-11e9-94ad-8afea6f69cc3.png">


Could you try removing your yarn.lock file .docz folder and node_modules and trying again ?

```sh
rm -rf .docz node_modules yarn.lock && yarn add docz@next && yarn docz build && yarn docz serve
```

mrfelton:
[...]

PR #1194, opened 2019-10-23
fix(docz-core): always copy theme under src (#1105)

simonrelet:
### Description

The theme directory should always be copied under _.docz/src/_ even when a custom path has been set using the `themesDir ` option.


simonrelet:
I locally tested using _dev-env/basic/_, should I also add an example under _example/_?

rakannimer:
Thanks for the PR ! 

An example folder similar to `basic` but called `with-themesDir` with a custom `themesDir` would be awesome. 

It would allow us to point people to a working example and eventually add the example to our test suite 👍

rakannimer:
Going to merge this to release it in `2.0.0-rc.64`, we can open a new PR for the example 👍

PR #1197, opened 2019-10-24
fix(docz): use playground from context (fixes playground shadowing)

rakannimer:
Remove docz-components and use Playground provided from gatsby-theme-docz via React Context


core/docz/src/components/Playground.tsx
```diff
@@ -14,38 +10,34 @@ export interface PlaygroundProps {
...
-      showLiveError,
-      showLivePreview = showLivePreviewFromUser,
-    },
-  } = useConfig()
+  const components = useComponents()
+  if (!components || !components.playground) return null
 
   return (
-    <DoczPlayground
+    <components.playground
```
Kerumen: Maybe we can create a variable for this no? Makes it a bit clearer:

```
const components = useComponents()
const PlaygroundComponent = components.playground
```

Also the first `!components` is not necessary as there is already a default value here: https://github.com/jesperorb/docz/blob/6eda1289c872f891f58ac6e51c94d66938884294/core/docz/src/hooks/useComponents.tsx#L51

examples/react-component/package.json
```diff
@@ -0,0 +1,10 @@
+{
```
Kerumen: Is it related?

rakannimer: Thanks for catching that :+1:

PR #1201, opened 2019-10-25
chore(docz-core): copy `.eslintignore` from project

Kerumen:
### Description

Gatsby has an Eslint plugin by default. It is run when we start the project. However, the cwd is at that time the `.docz` folder. Hence, if we add an `.eslintignore` in our project, it is not picked up. This add a build step to copy the file into the `.docz` folder if it exists.

Side note, I didn't understand what is the purpose of the function above, `writeEslintRc`, as it has an effect only in the Docz repository..


rakannimer:
Looks good ! 

I think `writeEslintRc` is because of the previous docz development setup that required the cli to behave differently when in the docz codebase.

PR #1203, opened 2019-10-25
style(gatsby-theme-docz): simplify the default config

Kerumen:
### Description

Due to the Playground move, some props were not used. I cleaned up a bit.


rakannimer (approved):
👍 !

PR #1211, opened 2019-10-28
style(gatsby-theme-docz): keep NavGroup opened on item click

Kerumen:
### Description

As all the menu items are now closed by default, if we open one and click on one item, it remount the component and closes it. This PR fixes that by keeping it opened.


rakannimer (approved):
Nice ! Didn't need additional state 👍

PR #1223, opened 2019-11-01
chore(gatsby-theme-docz): switch out react-helmet

selbekk:
This change switches out react-helmet with react-helmet-async. This
change removes an annoying warning about deprecated lifecycle usage, 
which makes at least my life much more enjoyable 😍


rakannimer:
Looks like react-helmet won't be fixing the warning on their side anytime soon, but I'm hesitant to merge this as it will break the setup for users who are using  `react-helmet` (without installing it) to set custom headers. 

If we're going to definitely make the switch then we'll also have to update the documentation telling users to `react-helmet-async` instead of `react-helmet`.

PR #1232, opened 2019-11-06
fix: broken dark mode screen caused by main html tag's height

marcelovicentegc:
### Description

Dark mode's current behavior is broken. Whenever its children overflow its parent current height (which is of `100vh`), white margins shows up. Changing the `main` tag style from `height: 100vh` to `minHeight: 100vh` fixes this, as the `main` tag will then obey its children height.

### Screenshots

| Before | After |
| ------ | ----- |
|Screenshot from 2019-11-06 13-55-56|Screenshot from 2019-11-06 13-58-28|


selbekk (approved):

rakannimer:
Thanks @marcelovicentegc !

PR #1245, opened 2019-11-11
Change the language of a warning

selbekk:
Noticed this little typo while re-building my components' props cache.


rakannimer:
Thanks !

Issue #1238, opened 2019-11-08
Custom document properties in useMenus() hook


NicolaiF:
At the start of every mdx-document, one has the ability to write custom properties for the chosen page. This is useful, but these properties, are only accessible through the useCurrentDoc() hook (as I can see). 

The feature request is to have these accessible through the useMenus() hook as well, so when implementing a menu for your site, one can customize how each `MenuItem` should appear based on these custom properties.

rakannimer:
Hey @NicolaiF,

Thanks for reporting this, I agree with you it would be a useful feature to have.

To enable it we'll need to delete the `entryAsMenu` function here : https://github.com/doczjs/docz/blob/master/core/docz/src/hooks/useMenus.ts#L12-L16

And remove the array mapping here : https://github.com/doczjs/docz/blob/master/core/docz/src/hooks/useMenus.ts#L19 and here : https://github.com/doczjs/docz/blob/master/core/docz/src/hooks/useMenus.ts#L29

And last, probably fix a type "error" here : https://github.com/doczjs/docz/blob/master/core/docz/src/hooks/useMenus.ts#L141

Would you like to give a PR a shot ? 

If not, no problem I can get to it later 👍

NicolaiF:
I will try it over the weekend, thanks for showing me the necessary steps needed!

rakannimer:
Awesome ! Feel free to join the docz slack workspace if you have any questions : https://join.slack.com/t/docz-workspace/shared_invite/enQtNzc4ODc2ODA3NzUxLWI0ZmYwY2NhNjhkNDFjN2UzYzlmMDcwZjQyZGY4MWQ1NmQwZDVlZDE4MmE3N2I4MWRjZTAxZjY4ODk4NGMzZjg

PR #1246, opened 2019-11-11
feat(docz): allow for custom properties in useMenus hook

NicolaiF:
### Description

Adding the feature of accessing custom properties from mdx-files in the useMenus hook.
As a side-effect of this, other preexisting fields such as slug and headings, are also included in the hook. This could potentially be improved by filtering these out, and only allowing for the original properties found in useMenus previously. 

### Pre-merge checklist

- [x] Check that all example projects are working.

### DIFF
--- a/core/docz/src/hooks/useMenus.ts
+++ b/core/docz/src/hooks/useMenus.ts
@@ -11,7 +11,7 @@ const noMenu = (entry: Entry) => !entry.menu
 const fromMenu = (menu: string) => (entry: Entry) => entry.menu === menu
 
 const entriesOfMenu = (menu: string, entries: Entry[]) =>
-  entries.filter(fromMenu(menu)).map(e => e as any) as MenuItem[]
+  entries.filter(fromMenu(menu))
 
 const parseMenu = (entries: Entry[]) => (name: string) => ({
   name,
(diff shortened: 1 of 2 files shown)


core/docz/src/hooks/useMenus.ts
```diff
@@ -9,14 +9,9 @@ import { Entry, MenuItem, doczState } from '../state'
...
 const fromMenu = (menu: string) => (entry: Entry) => entry.menu === menu
-const entryAsMenu = (entry: Entry) => ({
-  name: entry.name,
-  route: entry.route,
-  parent: entry.parent,
-})
 
 const entriesOfMenu = (menu: string, entries: Entry[]) =>
-  entries.filter(fromMenu(menu)).map(entryAsMenu)
+  entries.filter(fromMenu(menu)).map(e => e as any) as MenuItem[]
```
rakannimer: Could we remove the map since we're passing to it an identity function ?

rakannimer:
Hey @NicolaiF 

Thanks for the PR, looks great ! 

The tests will make sure to test the code against the examples so you shouldn't need to :+1: 

I added one small comment, that applies to both uses of `.map` with an identity function (e => e). 

Let me know what you think before we merge and publish.

NicolaiF:
Needed to update the types a couple of places to make it work without map, but it looks like everything works fine now!

rakannimer:
Awesome ! Thanks :)

rakannimer:
Fixes #1238 (for future reference)

PR #1251, opened 2019-11-13
fix(gatsby-theme-docz): ignore the .git folder

selbekk:
With this change, the git folder is ignored, improving the developer experience quite a bit :)


rakannimer:
Thanks @selbekk for catching and fixing this ! 👍

PR #1257, opened 2019-11-14
style: update Playground margin

mickaelzhang:
### Description

I updated the margin so that when we resize the playground, it stays on the left instead of in the center. It makes it easier to follow and I think this is what expected from the user.

### Review

Not sure if there is anything to review

### Screenshots

| Before | After |
| ------ | ----- |
| Playground before  | Playground|


rakannimer:
This is very cool ! Thanks @mickaelzhang :)

PR #1300, opened 2019-12-03
Update example to use correct Regexp

filipposarzana:
### Description

`[w-]*` means to match `w` character literally or `-` character, zero or unlimited times.

So now this Regex is matchingeverything that endsWith `.ts` or `.tsx`, even a file with empty filename.
If this is intended, so the first parti of the regular expression is completely useless. Otherwise we could be more strict on this (since we are speaking of an example, of course)

Moreover the capturing group around `(js|jsx|ts|tsx)` is not needed since it's unused inside a `.test` method and it's heavier than `(?:j|t)sx?

This new regex does exactly 1/4 of the steps than before.
*Before*: 85 steps
image

*After*: 21 steps
image

### Review

- [ ] Check the copy
- [ ] ...
- [ ] ...

### Pre-merge checklist

- [ ] ...
- [ ] ...

### Screenshots

| Before | After |
| ------ | ----- |
| image  | image |

### DIFF
--- a/examples/with-algolia-search/src/docs/general/usage-with-typescript.mdx
+++ b/examples/with-algolia-search/src/docs/general/usage-with-typescript.mdx
@@ -47,6 +47,6 @@ If you have particular needs or need to control exactly which files are being pi
 export default {
   filterComponents: files =>
     //This overrides the default filtering of components
-    files.filter(filepath => /[w-]*.(js|jsx|ts|tsx)$/.test(filepath)),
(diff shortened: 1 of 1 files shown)


rakannimer:
Hey @filipposarzana 

Thanks for the PR !

> [w-]* means to match w character literally or - character, zero or unlimited times.

Definitely an oversight, thanks for catching that and replacing it with a better expression 👍

PR #1304, opened 2019-12-06
Playground encapsulation / iframe around the playground

lkuechler:
### Description

This PR adds a iFrame around every playground to solve problems like these: https://github.com/doczjs/docz/issues/1271 https://github.com/doczjs/docz/issues/1170

One example for the above mentioned issues is the component in dev-env/basic (https://github.com/doczjs/docz/tree/master/dev-env/basic) which currently inherits the font-family from the surrounding docz styles.

I still have open todos for this PR but already wanted to created it to make sure that I am on the correct way for solving these issues. :)

### Drawbacks and issues

- any css-in-js library would need changes to still be able to apply styles
- existing projects could already depend on the provided css from docz (font-family in basic example)

### Open topics

- [ ] implement examples for other css-in-js libraries
- [ ] automatic resizing of the iframe would be helpful

Issue #1271, opened 2019-11-20
Prepend docz CSS classnames to prevent conflicts


auermi:
We have a component that uses the className `tag`. This is causing conflicts with elements with className of tag used in the Playground.

A potential solution would be to prepend all docz CSS classes with `docz-`

Example:
![Screen Shot 2019-11-20 at 10 01 12 AM](https://user-images.githubusercontent.com/5431160/69249914-b7757580-0b7c-11ea-98a3-4d96000103b5.png)

rakannimer:
Hey @auermi 

Thanks for reporting this issue. 

This seems [prism-react-renderer](https://github.com/FormidableLabs/prism-react-renderer) specific. 

We can either open an issue there with a request to allow us to prepend some string to the class names or wrap the playground editor in an iframe.

rakannimer:
Hey @auermi  

This is solved in 2.2.0. 

To scope your playground styles and make them independent of docz theme css you have 2 choices : 

1. add a useScoping prop to your playground

```jsx
<Playground useScoping> 
{
// blablabla
}
</Playground>
```
2. add `useScopingInPlayground : true` in doczrc.js 

```js
export default {
	useScopingInPlayground: true
}
```

Sorry for the delay in the reply but the fix was implemented thanks to you raising this issue ! So thanks 👍 

Going to close it, please re-open or comment if for some reason something doesn't work for you !

If you're interested in implementation details check #1305

auermi:
@rakannimer Thanks!

PR #1305, opened 2019-12-06
feat(gatsby-theme-docz): add optional iframe for preview and editor

rakannimer:
This PR adds an optional iframe around the editor and preview in Playground.

Still haven't decided if it should be opt-in or opt-out. 

Any thoughts on that would be appreciated 👍 

Similar PR #1304


core/gatsby-theme-docz/src/components/Playground/index.js
```diff
@@ -79,14 +99,22 @@ export const Playground = ({ code, scope, language }) => {
             </button>
           </div>
         </div>
+        {showingCode && (
+          <Wrapper height={editorHeight}>
```
lkuechler: Am i correct in the assumption that this was done to solve this issue #1271 ?

Because now that you are wrapping the component in a iframe the encapsulation of the editor should not be needed any more. Am i missing something? :)

rakannimer: > Am i correct in the assumption that this was done to solve this issue #1271 ? 
Yep that's exactly why ! Initially wanted to only wrap the preview.

I'm wrapping the preview and the editor each in an iframe, not wrapping the whole Playground with an iframe so it should be necessary if I'm not mistaken.

core/gatsby-theme-docz/src/components/Playground/IframeWrapper.js
```diff
@@ -0,0 +1,23 @@
+/** @jsx jsx */
+import { jsx } from 'theme-ui'
+import Iframe from 'react-frame-component'
+
+import * as styles from './styles'
+
+const CODE_FONT = `<style type="text/css">@import url(https://fonts.googleapis.com/css?family=Inconsolata&display=swap);</style>`
+const INITIAL_IFRAME_CONTENT = `<!DOCTYPE html><html><head> ${CODE_FONT} <style> \n body { padding: 0; margin: 0; }  </style></head><body><div></div></body></html>`
```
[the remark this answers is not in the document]
rakannimer: Yes, it would ! This felt a bit hacky, I'm not sure I like having the code_font styles and might remove them, because like you said it would go against what we're trying to do. 

Resetting the body margin and padding though shouldn't be too problematic I think. It's not necessary for sure but it felt ugly without it 😅 

I was trying to make the playground look exactly the same with or without the iframe but it's probably not a good idea.

lkuechler:
@rakannimer Is this the correct usage for this feature?

```
<Playground useIframe>
  <Alert>Some message</Alert>
</Playground>
```

rakannimer:
@lkuechler yep ! 

@Dangoo suggested to make it true by default but I'm not sure I'm convinced because this could be considered a breaking change. 

So to turn it off you would have to do  : 

```jsx
<Playground useIframe={false}>
  <Alert>Some message</Alert>
</Playground>
```

Btw feel free to join the [docz slack](https://join.slack.com/t/docz-workspace/shared_invite/enQtNzc4ODc2ODA3NzUxLWI0ZmYwY2NhNjhkNDFjN2UzYzlmMDcwZjQyZGY4MWQ1NmQwZDVlZDE4MmE3N2I4MWRjZTAxZjY4ODk4NGMzZjg) to join the conversation.

rakannimer:
- Renamed `useIframe` to `useScoping` (defaults to false)
- Added config flag `useScopingInPlayground` (defaults to false)

core/gatsby-theme-docz/src/components/Playground/index.js
lkuechler: I think the issue was that the css from the component inside the preview was styling the code area of the preview. This should not be possible any more after the component preview is wrapped in a iframe.

PR #1314, opened 2019-12-12
feat(docz-core): support kebab & pascal case component file name pattern

axe312ger:
### Description

Some of my projects are using kebab case as file name pattern. This adds support for pascal case file naming in case somebody feels like using it in their (legacy?) project.

### Review

- [ ] Should we add more patterns?


rakannimer:
Thanks for the PR @axe312ger !

CI is failing because of the absence of type defs for kebab-case and pascal-case. 

To fix it we can : 

1. replace the `kebab-case` module with kebabCase from lodash (`import { kebabCase } from 'lodash'`)
2. add @types/pascal-case as a dependency (`yarn add -D @types/pascal-case`)

What do you think ?

axe312ger:
I'd suggest switch all string case manipulations to run with the same framework.

As pascal/upper-camel-case is only supported by `change-case`, it covers all the other ones and exports every case manipulation function as a separate packe, we should got for that one:

* keep `pascalCase` and add typings
* keep `kebabCase` and add typings
* get rid of the capitalization one we have in other parts of the repo with `capitalCase`

What do you think?

axe312ger:
@rakannimer I added the types, still failing. Any idea?

rakannimer:
Can you try 
```js
import { pascalCase } from "pascal-case";
```

The error is : 

```
Error: src/components/Props.tsx:126:11 - error TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'typeof import("/home/circleci/repo/node_modules/pascal-case/dist/index")' has no compatible call signatures.

126       `/${pascalCase(componentName)}.`,
              ~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #1315, opened 2019-12-12
fix(docz-core): properly support node_modules for multiple file patterns

axe312ger:
### Description

Currently docs from node_modules are only supported when the src contains `node_modules`

This changes the code so that either src or any pattern can contain docs from `node_modules`

My config:
```js
{
  resolve: `gatsby-theme-docz`,
  options: {
    files: [
      `src/components/mdx/**/docs/*.{md,markdown,mdx}`,
      `node_modules/@gatsby-mdx-suite/*/docs/*.{md,markdown,mdx}`,
      `../../node_modules/@gatsby-mdx-suite/*/docs/*.{md,markdown,mdx}`,
    ]
  },
}
```

### DIFF
--- a/core/docz-core/src/lib/Entries.ts
+++ b/core/docz-core/src/lib/Entries.ts
@@ -47,14 +47,15 @@ export class Entries {
   }
 
   private async getMap(config: Config): Promise<EntryMap> {
-    const { paths, ignore, plugins, mdPlugins } = config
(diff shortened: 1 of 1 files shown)


rakannimer:
Awesome ! I added a src check to avoid breaking any setups that relied on it.

PR #1316, opened 2019-12-12
feat(docz-core): allow custom pattern for globbing components

axe312ger:
### Description

With #1315 I am able to include docs from within my dependencies and my source directory. This change adds support for globbing components from multiple directories.

My config:
```js
{
  resolve: `gatsby-theme-docz`,
  options: {
    files: [
      `src/components/mdx/**/docs/*.{md,markdown,mdx}`,
      `node_modules/@gatsby-mdx-suite/*/docs/*.{md,markdown,mdx}`,
      `../../node_modules/@gatsby-mdx-suite/*/docs/*.{md,markdown,mdx}`,
    ],
    customPattern: [
      `src/components/mdx/**/*.{js,jsx,mjs}`,
      `node_modules/@gatsby-mdx-suite/**/*.{js,jsx,mjs}`,
      `../../node_modules/@gatsby-mdx-suite/**/*.{js,jsx,mjs}`,
    ],
  },
}
```

### Review

- [x] is the naming correct?
- [x] should we merge the custom pattern with the default one?

### DIFF
--- a/core/docz-core/src/config/argv.ts
+++ b/core/docz-core/src/config/argv.ts
@@ -27,6 +27,7 @@ export interface DocgenConfig {
   resolver?: (ast: any, recast: any) => any
   propFilter?: (prop: any) => boolean
   searchPath: string
+  searchPatterns: string[]
 }
 
 export interface Menu {
--- a/core/docz-core/src/config/argv.ts
+++ b/core/docz-core/src/config/argv.ts
@@ -75,6 +75,7 @@ export interface Config extends Argv {
   themesDir: string
   mdxExtensions: string[]
   filterComponents: (files: string[]) => string[]
+  customPattern: string[]
 }
 
 export const setArgs = (yargs: Yargs) => {
--- a/core/docz-core/src/config/argv.ts
+++ b/core/docz-core/src/config/argv.ts
@@ -76,7 +76,6 @@ export interface Config extends Argv {
   themesDir: string
   mdxExtensions: string[]
   filterComponents: (files: string[]) => string[]
-  customPattern: string[]
 }
 
 export const setArgs = (yargs: Yargs) => {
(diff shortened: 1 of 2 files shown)


rakannimer:
Hey @axe312ger 

Thanks for the PR 👍 

I can definitely see the use of having an additional flag to give the user more control to tell docz where to look for components.

 > is the naming correct?

The name `customPattern` seems like it could be confusing. 

What do you think about adding it as a property to [`docgenConfig`](https://github.com/doczjs/docz/blob/master/core/docz-core/src/config/argv.ts#L26) called `searchPatterns`. 

>  should we merge the custom pattern with the default one?

We could add the logic to the [getPattern](https://github.com/doczjs/docz/blob/master/core/docz-core/src/states/props.ts#L12) function. If `docgenConfig.searchPatterns` is defined than we return it as is (we can assume that by providing this flag the user takes full responsibility of the search patterns and won't want us to tamper with it) and if not we'll keep the logic to return the patterns as is.

axe312ger:
Good idea! Ill refactor :)

axe312ger:
Done. I decided not to touch the user defined search pattern at all, because:

* only that way we give full control
* in some of my cases I want `node_modules` to be included
* just attaching `'!**/doczrc.js'` to the pattern felt odd as we can assume users will include their folder properly

Next todos:

* document the exact behaviour of how docz searches components when
  * using the default way, pattern is build out of root, src & docgen.searchPath + user ignores + default ignores
  * using `searchPatterns` which will give 100% control but you need to take care of where you looking for files #performance
* consider using `debug` module so users can see which files and folders are actually searched for docs & components by setting a environment variable (I feel like that be useful even without this pr!)

rakannimer:
Looks good ! 

I think the next todos you mentioned are very important and needed.

Let's merge this PR and we'll get to the next todos in different PRs.

Not sure if I've already invited you but please feel free to join [docz's slack workspace](https://join.slack.com/t/docz-workspace/shared_invite/enQtNzc4ODc2ODA3NzUxLWI0ZmYwY2NhNjhkNDFjN2UzYzlmMDcwZjQyZGY4MWQ1NmQwZDVlZDE4MmE3N2I4MWRjZTAxZjY4ODk4NGMzZjg) if you'd like to chat more about future features or issues.

PR #1319, opened 2019-12-13
fix(docz-core): forward cli status code properly

adbayb:
### Description

Currently, spawned processes (for example, `gatsby build`) run through CLI don't forward error code correctly to the parent process (`docz build`) producing success code (eg. `0`) even if some child processes failed. 
It can be problematic especially if we rely on it to do some CI/CD (like building a Docker image).
This PR solves the issue by forwarding properly child error code on `build` and `serve` commands.

### Screenshots

See `echo $?` output.

| Before | After |
| ------ | ----- |
| Capture d’écran 2019-12-13 à 17 19 50  | Capture d’écran 2019-12-13 à 17 15 50 |

### DIFF
--- a/core/docz-core/src/bundler/build.ts
+++ b/core/docz-core/src/bundler/build.ts
@@ -1,11 +1,11 @@
 import * as fs from 'fs-extra'
 import * as path from 'path'
-import spawn from 'cross-spawn'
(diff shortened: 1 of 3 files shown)


rakannimer:
This is very helpful ! Thanks 😄 !

PR #1332, opened 2019-12-17
fix(docz-core): prevent false positive CLI error code

adbayb:
### Description

Following this PR https://github.com/doczjs/docz/pull/1319, I noticed that npm could use `stderr` file descriptor to write warning messages (which are not considered as error). Consecuently, we could have in this case false positives error code like this (please focus on status value and stderr output):

Capture d’écran 2019-12-17 à 11 53 39

Relying on `spawn.status` exit success value is more reliable.


rakannimer:
Makes sense, nice catch 👍

PR #1368, opened 2020-01-27
fix(gatsby-theme-docz): removed unused gatsby-plugin-manifest

alchi126:
### Description

`gatsby-theme-docz` has a dependency on `gatsby-plugin-manifest`, which in turn relies on a package called `sharp` that's causes install issues on Windows x64 machines. 

As `gatsby-plugin-manifest` serves a specific use case (see specs [here](https://www.gatsbyjs.org/packages/gatsby-plugin-manifest/)) and is actually never used in `gatsby-theme-docz`, it'd be a great help to remove it from the theme all together.

### DIFF
--- a/core/gatsby-theme-docz/package.json
+++ b/core/gatsby-theme-docz/package.json
@@ -30,7 +30,6 @@
     "gatsby-plugin-alias-imports": "^1.0.5",
     "gatsby-plugin-compile-es6-packages": "^2.0.0",
     "gatsby-plugin-emotion": "^4.1.2",
-    "gatsby-plugin-manifest": "^2.2.3",
     "gatsby-plugin-mdx": "^1.0.13",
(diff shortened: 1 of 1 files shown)


rakannimer:
Good catch ! Thanks 👍

alchi126:
Thanks for the merge, and all the great work you've been doing with docz!

alchi126:
When is the next release expected to be? Due to external download restrictions on my work machine I could only get the package via npm. 

On a related note: `docz` also has an unused dependency on `gatsby-theme-docz`. I don't think it'd cause any issues, providing the version gets bumped after the next release of `gatsby-theme-docz`, but I'll gladly submit a PR to remove if it helps.

wj2061:
I can't use docz@2.2.0  for the same reason , now I have to use  docz@2.3.0-alpha.13 instead.

PR #1381, opened 2020-02-07
feat(gatsby-theme-docz): create MainContainer component

mickaelzhang:
### Description

There were no way to target the main container, where the content is. With this, you can shadow this component without impacting any other component.


rakannimer (approved):

rakannimer:
👍 Released in v2.3.0-alpha.11 or `docz@next`

Issue #1359, opened 2020-01-13
Light / Dark mode button is not accessible 


rakelbjort:
## Bug Report

**Describe the bug**
The button in the upper right corner for switching between light or dark mode for the application is not accessible for screen readers

**Expected behavior**

Please consider adding an aria-label on the button.

Suggestion "Switch to dark mode" / "Switch to light mode"

PR #1383, opened 2020-02-09
[gatsby-theme-docz]: add aria-label attribute to theme button

alexandre-lelain:
issue: https://github.com/doczjs/docz/issues/1359

### Description

This PR adds an `aria-label` attribute on the Header's button responsible for switching between light and dark mode for screen readers.

### Review

- [ ] Check that there is an `aria-label="Switch to light|dark mode"` attribute on the theme button in the Header.


rakannimer:
This is cool, thanks @alexandre-lelain !

Issue #1374, opened 2020-01-31
How can I disable the --open browser?


hbrls:
## Question

**Description**
 
The `--open` is a cli option of gatsby. How can I disable it?

`docz dev --open=false` takes no effect.

`docz-config.js openBrowser: false` takes no effect.

rakannimer:
Hey @hbrls,

Thanks for reporting this, it should be fixed in `2.3.0-alpha.11`, you can try it now with `docz@next` or in `2.3.0` when it is released 👍

PR #1384, opened 2020-02-09
fix(docz-core): pass down host when running serve

ShinyLeee:
### Description

When I start deploy docz project in my own server, i found that server publish ip address can not be access but localhost can, and after checkout [gatsby cli command](https://www.gatsbyjs.org/docs/gatsby-cli/#serve), i found the problem is host is always be `localhost` and can not be changed to others when use `docz serve`.

### DIFF
--- a/core/docz-core/src/commands/serve.ts
+++ b/core/docz-core/src/commands/serve.ts
@@ -11,10 +11,17 @@ export const serve = async (args: Arguments<any>) => {
 
   if (config.port) {
     cliArgs.push('--')
-    // Append gatsby option `port`to CLI args
+    // Append gatsby option `port` to CLI args
     // https://www.gatsbyjs.org/docs/cheat-sheet/#cheat_sheet-text
     cliArgs.push('--port')
     cliArgs.push(String(config.port))
+  }
+
+  if (config.host) {
+    // Append gatsby option `host` to CLI args
+    // https://www.gatsbyjs.org/docs/cheat-sheet/#cheat_sheet-text
+    cliArgs.push('--host')
+    cliArgs.push(String(config.host))
   }
 
   sh.cd(paths.docz)


core/docz-core/src/commands/serve.ts
```diff
@@ -11,12 +11,19 @@ export const serve = async (args: Arguments<any>) => {
...
   if (config.port) {
     cliArgs.push('--')
-    // Append gatsby option `port`to CLI args
+    // Append gatsby option `port` to CLI args
     // https://www.gatsbyjs.org/docs/cheat-sheet/#cheat_sheet-text
     cliArgs.push('--port')
     cliArgs.push(String(config.port))
   }
 
+  if (config.host) {
```
alexandre-lelain: I think we should do a more generic gatsby arguments handling, so users can use all of gatsby's cli arguments.

Something like:

```js
for (const arg in config) {
  cliArgs.push('--')
  cliArgs.push(`--${arg}`)
  cliArgs.push(String(arg))
}
```

This code will make sure to scale automatically when gatsby are adding new cli arguments.
Also, this would probably solve https://github.com/doczjs/docz/issues/1374.

What do you think @rakannimer ?

ShinyLeee: config object now contains project config (menu, themeConfig...) other than gatsby's cli arguments.

image

If we want make it scale automatically, maybe add a `gatsbyCliConfig` to place these gatsby's arguments?

rakannimer: Hey @alexandre-lelain, what you're saying makes sense to me, and I tried doing something like that a couple of commits ago but it resulted in some errors because of conflicts over variable names between docz config and gatsby. So while it looks redundant it might be better (for now) to explicitly pass down arguments.

Regarding #1374 it should be fixed in `2.3.0-alpha.11` but I forgot to update so thanks for the reminder !

rakannimer:
Thanks @ShinyLeee 👍

PR #1389, opened 2020-02-12
fix(gatsby-theme-docz): use sx instead of style prop in Playground

floriangosse:
### Description

Fixes the missing padding for in the preview of the `Playground` component.


rakannimer:
Awesome ! Thanks 👍

rakannimer:
Released in `next` as version : `2.3.0-alpha.12`

PR #1415, opened 2020-03-10
fix(docz-core): always use custom prop filter if it's provided (#1403)

### LINKED ISSUE #1403: docgenConfig.propFilter is ignored
## Bug Report

I'm passing a custom `docgenConfig.propFilter` to my `docz.config.js`, but it is not being run.

Culprit seems to be [here](https://github.com/doczjs/docz/blob/87eea505e3a9aef78c53b6bdcdc39d596d37bb7c/core/docz-core/src/utils/docgen/typescript.ts#L170) in the TypeScript docgen parser module.

It's short, so I'll copy it here for convenience:
```ts
propFilter(prop: any): any {
  if (prop.parent == null) return true
  const propFilter = config.docgenConfig.propFilter
  const val = propFilter && _.isFunction(propFilter) && propFilter(prop)
  return !prop.parent.fileName.includes('node_modules') || Boolean(val)
}
```

This code ignores the custom `propFilter` if `prop.parent == null`, but I would like to have my custom `propFilter` always run so that I can decide how to treat `prop.parent == null`.

Also, in the return statement, my custom `propFilter` result could also be ignored. It seems to say that if the prop's parent is not a node_modules file, then it should always be included regardless of my `propFilter` result.

## Suggestion

Provide a "default" propFilter, but allow the user to completely override it if they want.

```js
function defaultPropFilter(prop) {
  if (prop.parent == null) return true;
  return !prop.parent.fileName.includes('node_modules');
}

 propFilter(prop: any): any {
    const customPropFilter = config.docgenConfig.propFilter;
    const propFilter =_isFunction(customPropFilter) ? customPropFilter : defaultPropFilter;
    return Boolean(propFilter(prop));
  }
```

**Environment**

- docz version: 2.2.0
- OS: OSX 10.15.3
- Node/npm version: Node 12.14.1 / NPM 6.13.4

tripphamm:
### Description

Closes #1403 

I'm curious how y'all feel about considering this a "breaking change" vs just a "fix".

Technically the new behavior is different from the old behavior and folks who are using a custom propFilter might end up seeing more props after this is merged.

Here's a summary:

**Old behavior (without a custom prop filter):**
- any props that don't have a `parent` are included
- any props that have a non-node_module `parent` are included

**New behavior (without a custom prop filter):**
- same as old behavior

---

**Old behavior (with custom prop filter):**
- any props that don't have a `parent` are included
- any props that have a non-node_module `parent` are include
- any props that 1) have a `parent` and 2) `parent` is a node_module - will be run through the custom prop filter

This behavior seemed like a bug to me. The custom prop filter is almost always ignored and is only useful for white-listing some props that come from node_modules

**New behavior (with custom prop filter):**
- all props are run through the custom prop filter

So, if folks were actually relying on the old "buggy" behavior of propFilter just being a node_modules whitelist, they will need to add some logic to their propFilter in order to re-create the old behavior.

Overall I believe that this new API is the _right_ API, and it's just a matter of how to approach the change with the community, which I'll leave up to the maintainers

### DIFF
--- a/core/docz-core/src/utils/docgen/typescript.ts
+++ b/core/docz-core/src/utils/docgen/typescript.ts
@@ -165,13 +165,20 @@ function createServiceHost(
   }
 }
 
+const defaultPropFilter = (prop: any): boolean => {
+  return prop.parent == null || !prop.parent.fileName.includes('node_modules')
+}
+
 const parseFiles = (files: string[], config: Config, tsconfig: string) => {
   const opts = {
-    propFilter(prop: any): any {
-      if (prop.parent == null) return true
-      const propFilter = config.docgenConfig.propFilter
-      const val = propFilter && _.isFunction(propFilter) && propFilter(prop)
-      return !prop.parent.fileName.includes('node_modules') || Boolean(val)
+    propFilter(prop: any): boolean {
+      const customPropFilter = config.docgenConfig.propFilter
+      const propFilter =
+        customPropFilter && _.isFunction(customPropFilter)
+          ? customPropFilter
+          : defaultPropFilter
+
+      return Boolean(propFilter(prop))
     },
     componentNameResolver(exp: ts.Symbol, source: ts.SourceFile): any {
       const componentNameResolver = config.docgenConfig.resolver


tripphamm:
CI is failing here:

```
/examples/typescript/src/components/Button.tsx
  53:3  error  Parsing error: Unexpected token

  51 | export interface ButtonProps {
  52 |   scale: 'small' | 'normal' | 'big'
> 53 |   kind: 'primary' | 'secondary' | 'cancel' | 'dark' | 'gray'
     |   ^
  54 |   outline: boolean
  55 | }
  56 | 
```

This doesn't seem like it's related to my change.

I saw this error during development as well. Curiously, adding semi-colons fixes the issue, but I wouldn't expect that they're needed.

rakannimer:
Can you rebase from master, I've fixed the CI issue

tripphamm:
Looks good now 👍

rakannimer:
Thank you !

PR #1417, opened 2020-03-13
fix(babel-plugin-export-metadata): add case for export default memo()

yoo2001818:
### Description

babel-plugin-export-metadata currently doesn't handle using call expression inside export default, such as:
```jsx
export default React.memo(Component)
```

This means it won't include `__filemeta` in the default export, so `<Props>` won't work correctly because it can't resolve the file path.

It resolves such case by separating expression and `export default`, to allow injecting `__filemeta` correctly.

```jsx
const __DOCZ_DUMMY_EXPORT_DEFAULT = React.memo(Component)
export default __DOCZ_DUMMY_EXPORT_DEFAULT
```

### DIFF
--- a/other-packages/babel-plugin-export-metadata/src/index.js
+++ b/other-packages/babel-plugin-export-metadata/src/index.js
@@ -19,6 +19,11 @@ const replaceExportDefault = template(`
   export default NAME
 `)
 
+const replaceExportDefaultCall = template(`
+  const NAME = DECLARATION
+  export default NAME
+`)
+
 const getFilename = state => {
   const rootDir = get(state, 'opts.root', process.cwd())
   const filename = get(state, 'file.opts.filename')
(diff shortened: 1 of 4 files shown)


yoo2001818:
I wasn't able to reproduce this problem in development build, probably because hot loader already rewrites code like this. However, production build doesn't include `__filemeta`.

Before https://github.com/doczjs/docz/commit/9d5b12983c1f40ffd3bfa5c3debb4ca8448db3fc#diff-5cc2196461930cb24939fb44764d656b, lack of __filemeta simply ended up not showing Props (Because `componentName` was `null`). However, after that commit, it broke production build - it threw an error.

The reason was somewhat weird - `pascal-case` package throws an error when `null` is passed. At first, I tried to resolve the problem by adding simple null check like `pascalCase(componentName || '')`, but I thought it was better to fix the fundamental problem.

rakannimer:
Thanks @yoo2001818 for the PR !

PR #1437, opened 2020-03-30
fix(gatsby-theme-docz): content overflowing issue

mickaelzhang:
### Description
In some cases the main container was overflowing, so a horizontal scroll bar was appearing which isn't what we expect. After digging a bit, this seems to solve the issue (related article: https://css-tricks.com/preventing-a-grid-blowout/)


rakannimer:
Looks good ! Thanks @mickaelzhang

PR #1468, opened 2020-04-23
feat(docz): bump gatsby dependencies to 2.20.29

dan2k3k4:
Bumping Gatsby dependencies to ^2.20.29 for Gatsby Recipes support.

Depends on #1487 releasing a newer version of the eslint config


jesperorb:
Seems like upping the deps for gatsby results in upping some babel-stuff which seems to result in a new version av `eslint`-plugins which makes the build fail.

PR #1477, opened 2020-04-29
feat(gatsby-theme-docz): configure SEO information from md/mdx files

shkoliar:
### Description

Change adds the possibility to set SEO meta information directly from *.md or *.mdx files without shadowing components.

For example:
```
---
title: Page1 Title
description: Description text
keywords: [keyword1, keyword1]
meta: []

name: Page1
route: /page1
---
```

Configuring SEO information is on a per-page basis. 

The page title can be customized and not depend on the page name. 

If this extra information is not provided in *.md or *.mdx file, the page name is used as a fallback.


jesperorb:
Seems reasonable, any reason why we would want to filter the frontmatter before passing to `SEO`/`Helmet` @rakannimer?

PR #1588, opened 2020-11-17
Code Updates

stramel:
### Description
Below is a detailed list of the changes:

#### Swap `SFC` for `FC`
The reasoning behind this is that `SFC` is marked as deprecated by the React team with the suggestion to use `FunctionComponent`/`FC` instead. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30364 

image

#### Update `react` imports to prepare for the new JSX transform
With React v17, the React team, launched a new JSX transform that doesn't require `React` to be imported or in the file. This will also be available in v16.14.0 https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

Based on the above announcement, the React team provides steps to take to swap over to the new JSX transform. The `default` export should no longer be used and use the named exports instead. (**NOTE:** You can still use the namespace import if you wish) This should make it easier to switch over in the near future.

I attempted to retain some of the import styling that I had seen throughout. (ie. Type imports as a separate import)

#### Update React `Fragment` usages
As of v16.2.0, Fragments have shorthand syntax available. https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

I swapped over all usages of `<React.Fragment>` to `<>`. NOTE: I used `<Fragment>` (imported from the named export, refer to above section) when we were utilizing `theme-ui`'s `jsx` pragma. 

#### Other minor changes
I noticed a few minor things that I fixed along the way:

- Removed a few `eslint-disable`s that were unused
- Added a few named function components (**NOTE:** Some of these were to satisfy a false positive with `react/display-name` eslint rule and others were to follow best practice and better work with the FastRefresh HMR)
- Prettier formatting

**NOTE:** I'm happy to revert/address any comments/concerns related to the changes made in this PR.

### DIFF
--- a/core/docz-core/__fixtures__/Alert/Alert.tsx
+++ b/core/docz-core/__fixtures__/Alert/Alert.tsx
@@ -1,4 +1,4 @@
-import React, { SFC } from 'react'
+import React, { FC } from 'react'
 import styled from '@emotion/styled'
 
 export type Kind = 'info' | 'positive' | 'negative' | 'warning'
@@ -27,6 +27,6 @@ const AlertStyled = styled('div')<AlertProps>`
   background: ${({ kind = 'info' }) => kinds[kind]};
 `
 
-export const Alert: SFC<AlertProps> = ({ kind, ...props }) => (
+export const Alert: FC<AlertProps> = ({ kind, ...props }) => (
   <AlertStyled {...props} kind={kind} />
 )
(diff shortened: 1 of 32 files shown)


pedronauck:
Thanks dude, PR accepted 🙏🏻

pedronauck:
hey @stramel I'll talk with you here because your DM on Twitter is closed 😂 please, send me a message, I have some proposal to talk with you that maybe you can have interesting 🙏🏻

PR #1689, opened 2022-01-07
Style: Inherit border radius in live editor

BretCameron:
### Description

The live editor has rounded edges, but the child `div` (which has a background) does not inherit the `border-radius`, leading to imperfect edges!

### Screenshots

Focus on the bottom left corner of each image!

| Before | After |
| ----- | ----- |
|  Screenshot 2022-01-07 at 11 42 56 | Screenshot 2022-01-07 at 11 42 45  |


renatobenks (commented):
I see the reason, well caught

renatobenks (approved):

Issue #1662, opened 2021-10-11
Support react 17


Bajdzis:
Currently, if we install the docz, we will get an error:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: test@1.0.0
npm ERR! Found: react@17.0.2
npm ERR! node_modules/react
npm ERR!   react@"^17.0.2" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.8.0" from docz@2.3.1
npm ERR! node_modules/docz
npm ERR!   docz@"*" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.

```

designbyadrian:
For those who encounter the error and don't know how to proceed, the temp fix is to bypass the peer installation effort:

`npm install docz --save --legacy-peer-deps`

This tells the installer to ignore trying to install an older version of react and react-dom on top of the recent version your project has.

But it's likely you're running a later version of Node as well, so this issue will probably be the next thing you'd have to look at: "ERR_REQUIRE_ESM error on start" https://github.com/doczjs/docz/issues/1674

prma85:
Created a PR to fix it. Let wait for the mainteners
https://github.com/doczjs/docz/pull/1692

renatobenks:
fixed by #1692

PR #1692, opened 2022-01-13
Fix: Support react 17 Issue #1662

prma85:
### Description

Add React 17 to peer dependencies to not break when used within an updated project

### Review

- [X] Check the copy
- [X] Build package and install with yarn link

### Pre-merge checklist

- [X] Check for versions

### Screenshots

Not needed

### DIFF
--- a/core/docz/package.json
+++ b/core/docz/package.json
@@ -1,6 +1,6 @@
 {
   "name": "docz",
-  "version": "2.3.3-alpha.0",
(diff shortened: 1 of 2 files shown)


renatobenks (approved):
Great! Thanks for the contribution. We actually have done this in the upcoming major version, in a different way, but I guess it's totally fine we move this way sooner to get it released sooner

PR #1696, opened 2022-01-31
fix(docz-utils): fix closing tag detection in `removeTags`

### LINKED ISSUE #1691: SyntaxError: Expected corresponding JSX closing tag
Getting `SyntaxError: Expected corresponding JSX closing tag` error when using any component with children inside the `Playground` component.

Using components that self close, for example `<MyComponent />` works fine.

Here is a sample with the error. Please click on Components > Alert on the left menu.

https://codesandbox.io/s/docz-example-xj522

Thanks

nicolo-ribaudo:
### Description

I wrote the description of this fix and the reason behind it at https://github.com/doczjs/docz/issues/1691#issuecomment-1025928444.

I fixed it in different commits:
- `test(docz-utils): add test for current removeTags behavior` shows the old behavior (one test was correct, the other one was already broken)
- `chore(docz-utils): update @babel/* dependencies` updates `@babel/traverse` and the other babel packages, to force the version containing the fix for the bug docz was relying on
- `test(docz-utils): update removeTags (broken) snapshots` updates the tests to show the new broken behavior
- `fix(docz-utils): fix closing tag detection in removeTags` fixes both the broken tests

Fixes #1691

### DIFF
--- a/core/docz-utils/.babelrc
+++ b/core/docz-utils/.babelrc
@@ -1,0 +1,14 @@
+{
+  "plugins": ["lodash"],
+  "presets": [
+    "@babel/preset-typescript",
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "current"
+        }
+      }
+    ]
+  ]
+}
--- a/core/docz-utils/src/jsx.ts
+++ b/core/docz-utils/src/jsx.ts
@@ -2,7 +2,7 @@ import * as jsxUtils from 'jsx-ast-utils'
 import strip from 'strip-indent'
 import escapeJS from 'js-string-escape'
 
-import { valueFromTraverse, codeFromNode } from './ast'
+import { valueFromTraverse } from './ast'
 
 export const propFromElement = (prop: string) =>
   valueFromTraverse(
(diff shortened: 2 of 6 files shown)


JLHwung (approved):

adbayb (approved):
Thank you @nicolo-ribaudo . 
It seems to work properly 🎉 (tested in my project via [patch-package](https://github.com/ds300/patch-package)).

core/docz-utils/src/jsx.ts
```diff
@@ -2,19 +2,29 @@ import * as jsxUtils from 'jsx-ast-utils'
...
-export const removeTags = (code: string) => {
-  const open = codeFromNode(p => p.isJSXOpeningElement())
-  const close = codeFromNode(p => p.isJSXClosingElement())
+const getTagContentsRange = valueFromTraverse<[number, number] | null>(
+  p => p.isJSXElement(),
+  ({ node }) => {
+    if (!node.closingElement) {
+      // if the JSX element doesn't have a closingElement, it's because it's self-closed
+      // and thus does not have any content: <Playground />
+      return null
```
renatobenks, line 19: what about returning `[0, 0]` as the default value?

core/docz-utils/.babelrc
```diff
@@ -0,0 +1,14 @@
...
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "current"
+        }
+      }
+    ]
+  ]
+}
```
renatobenks, line 14: why did you have to add it? it shouldn't be needed

renatobenks:
Great job at this @nicolo-ribaudo 🎉

core/docz-utils/src/jsx.ts
nicolo-ribaudo: With invalid inputs (i.e. if it can't find a JSX element) the function returned by `valueFromTraverse` returns an empty string, so we still need the default value fallback when calling `getTagContentsRange` on line 26.
- I don't think this will happen in practice, but it's better to be safe and avoid crashing
- Otherwise Typescript complaints

Since we need the default value there anyway, returning something "falsy" here allows us to specify the default value only once.

core/docz-utils/.babelrc
nicolo-ribaudo: Without this Jest was crashing on the TS syntax: it was not configured yet just because this package didn't have any tests. I copied it from other packages in this repo which already had some tests.

Elgeneinar:
Any estimate on when this will be completed?

core/docz-utils/.babelrc
renatobenks: I thought that probably it would be the reason, but thanks for explaining

core/docz-utils/src/jsx.ts
renatobenks: interesting to know that, it's fair enough! well done btw

renatobenks (approved):

renatobenks:
@Elgeneinar I'm scheduling the release to be done between today and tomorrow, but I'll try to let everyone knows in all following corresponded PRs

PR #1749, opened 2023-09-08
Added peerDependencies to support latest react version

shivam-pawar:
### Description

Added peerDependencies to support latest react version.


Monceee:
Sab, 9 Sep 2023 pukul 01.55 Shivam Pawar ***@***.***> menulis:

